PR #21, opened 2017-11-28
Added methods to load the image collection and set foreground/background

astrofrog:
@ojustino - just so you are aware, I've added a few methods related to the foreground/background. With this, you can see what imagery is available by checking ``wwt.available_layers`` and then set the foreground/background with e.g. ``wwt.foreground = ...``


astrofrog:
Oops I messed up this PR, will open a new one (this wasn't really merged)

PR #24, opened 2017-11-29
Initial try fixing core.py initialization bug

ojustino:
On [line 28](https://github.com/ojustino/pywwt-web/blob/force-traits/pywwt_web/core.py#L28), `_on_trait_change()` calls [WWTQtWidget's `_send_msg()` method from qt_widget.py](https://github.com/ojustino/pywwt-web/blob/force-traits/pywwt_web/qt_widget.py#L140). However, when using the new [`for` loop on line 14](https://github.com/ojustino/pywwt-web/blob/force-traits/pywwt_web/core.py#L15), I get `AttributeError: 'WWTQtWidget' object has no attribute 'widget'`. For some reason it's not accessing the definition for `self.widget` [in the WWTQtWidget class' `__init__()` method from qt_widget.py](https://github.com/ojustino/pywwt-web/blob/force-traits/pywwt_web/qt_widget.py#L125) and I can't figure out why.

### DIFF
--- a/pywwt_web/core.py
+++ b/pywwt_web/core.py
@@ -11,12 +11,14 @@ DEFAULT_SURVEYS_URL = 'http://www.worldwidetelescope.org/wwtweb/catalog.aspx?W=s
 
 
 class BaseWWTWidget(HasTraits):
-
+    
     def __init__(self):
-        super(BaseWWTWidget, self).__init__()
+        super(BaseWWTWidget, self).__init__() # does super() alone do the same?
         self.observe(self._on_trait_change, type='change')
         self._available_layers = []
         self.load_image_collection(DEFAULT_SURVEYS_URL)
+        for name in self.trait_names():
+            self._on_trait_change({'name': name, 'new': getattr(self,name), 'type': 'change'})
 
     def _on_trait_change(self, changed):
         # This method gets called anytime a trait gets changed. Since this class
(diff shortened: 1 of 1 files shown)


pywwt_web/core.py
```diff
@@ -8,19 +8,23 @@
 
 
 class BaseWWTWidget(HasTraits):
-
+    
     def __init__(self):
-        super(BaseWWTWidget, self).__init__()
+        super(BaseWWTWidget, self).__init__() # does super() alone do the same?
```
astrofrog: In Python 3 we could replace ``super(BaseWWTWidget, self)`` by ``super()`` but we need to be Python 2-compatible

ojustino:
The fix worked, but can you explain in more depth why putting the `super()` call first blocked `_send_msg()` from accessing the rest of the **init** in `qt_widget.py`?

astrofrog:
@ojustino - basically the code you added in ``core`` calls ``_on_trait_change`` which triggers a call to ``_send_msg``:

https://github.com/WorldWideTelescope/pywwt/blob/master/pywwt_web/core.py#L29

which then calls here:

https://github.com/WorldWideTelescope/pywwt/blob/master/pywwt_web/qt_widget.py#L144

However, that method needs ``self.widget`` to exist. However, if ``BaseWWTWidget.__init__`` is called first in ``WWTQtWidget`` then ``self.widget``, which is set here:

https://github.com/WorldWideTelescope/pywwt/pull/24/files#diff-f65245291bf7c921c97edb8c35a29cacR132

has not been set up yet. Does this make sense?

ojustino:
I think so. This means it's always necessary to put `super()` after any elements of a parent's *init* that a child needs to access?

astrofrog:
@ojustino - this may be quite a special case in terms of the order of the ``__init__``, it's not very common to have the __init__ of the base class call a method on the subclass that requires the subclass __init__ to have run

Issue #23, opened 2017-11-28
Expose option to control units of circle radius


astrofrog:
Currently the radius of the circle annotations is in display units but I think there is a way to specify it in arcsec (need to check WWT SDK docs)

ojustino:
The docs say that if [skyRelative](https://worldwidetelescope.gitbooks.io/worldwide-telescope-web-control-script-reference/content/webcontrolobjects.html#circle-skyrelative-property) is `True`. Does fixing this involve anything more than changing `skyRelative`'s default value to `True`, then?

astrofrog:
Done in #26

PR #26, opened 2017-11-30
Changing sky_relative default in core.py [issue #23]

ojustino:


pywwt_web/annotation.py
```diff
@@ -52,7 +52,7 @@ class Circle(Annotation):
     line_color = Unicode('white', help='Assigns line color for the circle').tag(wwt='lineColor', sync=True)
     line_width = Float(1, help='Assigns line width (in pixels)').tag(wwt='linewidth', sync=True)
     radius     = Float(10, help='Sets the radius for the circle').tag(wwt='radius', sync=True)
-    sky_relative = Bool(False, help='Whether the size of the circle is absolute or relative').tag(wwt='skyRelative', sync=True)
+    sky_relative = Bool(True, help='Whether the size of the circle is absolute (in arcseconds) or relative').tag(wwt='skyRelative', sync=True)
```
astrofrog: I've looked at this a little and I think we could do something more user-friendly, since we don't need to match exactly the WWT JS API. Maybe we could make it so that users can do:

```
from astropy import units as u
circle.radius = 3 * u.arcsec
circle.radius = 5 * u.pixel
```

and based on the units passed to radius, we adjust skyRelative automatically. Maybe this logic can be done inside ``_on_trait_change``. Maybe the Circle class can define its own ``_on_trait_change``, kind of like this (pseudo code, not tested)

```
def _on_trait_change(self, changed):
    if changed['name'] == 'radius':
        if changed['new'].unit.is_equivalent(u.pixel):
            self._send_msg(message here to change skyRelative to True)
            changed['new'] = changed['new'].to(u.pixel).value  # this removes the units
        else:
            self._send_msg(message here to change skyRelative to False)
            changed['new'] = changed['new'].to(u.arcsec).value
    super(Circle, self)._on_trait_changed(changed)
```

Does this make sense?

ojustino: I think so. I'll try this out and see how it goes.

ojustino: Is there a quick way to print an error message in the interpreter in JS for `wwt_json_api.js` if an invalid action is called for by the Python code? I'm trying to troubleshoot.

astrofrog: @ojustino - try using ``console.log`` (use it like print in Python). For the Jupyter version it will print in the JS console.

ojustino: Is there a way to do this through qt? I'm having problems running Jupyter at all on my setup.

astrofrog: I think it should work with the Qt version, it will just print the output into the terminal where you launched the Qt widget from.

ojustino: That and the other output methods I've found from searching haven't worked. The best I can manage is to print the final output sent from Python in `_send_msg()` from `qt_widget.py`, but the JavaScript fails/succeeds silently.

astrofrog: @ojustino - just for info I found/fixed the issue inhttps://github.com/WorldWideTelescope/pywwt/pull/30

ojustino:
The changes made here seem to work for the most part. I also changed the line width types for Poly() and PolyLine() objects from float to `AstropyQuantity` and gave them their own version of `_on_trait_change()` as well. A couple of questions:

1) Is just changing self.sky_relative depending on the unit of the radius OK? The commented `_send_msg()` doesn't change the unit as it should.

2) I commented out some print statements that I was using for troubleshooting. Is it OK to leave them for future use, or would you prefer them to be left out of this pull request?

astrofrog:
@ojustino - thanks for getting this to work!

1. yes we should just change depending on the units of the radius, but do you mean that this doesn't work currently if you uncomment ``_send_msg``?

2. we should ideally remove print statements to keep the 'real' code clean

pywwt/annotation.py
```diff
@@ -52,16 +56,42 @@ class Circle(Annotation):
...
                               dec=coord_icrs.dec.degree)
 
+    def _on_trait_change(self, changed):
+        #print('new _otc: ', changed, sep=' ')
+        if changed['name'] == 'radius':
+            if changed['new'].unit.is_equivalent(u.pixel):
+                # self.parent._send_msg(event='annotation_set',
+                #                       id=self.id,
+                #                       setting='sky_relative',
+                #                       value='true')
```
astrofrog: I think this should be ``value=True`` (not in quotes). Same for false below.

ojustino:
1) Nothing breaks, it's just that `sky_relative` and the units don't change. So if I start in pixels and want to switch to arcsec with `_send_msg()` enabled, I have to manually change `circle.sky_relative`. With the current setup, if I switch units, `sky_relative` will automatically switch as well. I tried with `True` and `False` earlier and the same happened; I was just testing different ways of sending the message.

ojustino:
All good now, the problem was the discrepancy between the WWT Web Control name (`skyRelative`) and its adapted Python name (`sky_relative`).

astrofrog:
@ojustino - looks good! Before I merge, I was wondering whether for ``line_width`` we should just use a plain float - for the radius it makes sense to require a unit since there is an ambiguity between pixels and angle units, but for line width since there isn't a choice, maybe we don't need to do anything with units in that case?

ojustino:
We could. I changed `line_width` because it was the only trait that didn't take an astropy quantity (among those that take numbers of some sort), which I thought didn't fit.

astrofrog:
Ok, I'll merge it as-is and we can always revisit later if needed - thanks!

Issue #4, opened 2017-11-14
Add methods to load, play, and stop tours


astrofrog:
There are methods here:

https://worldwidetelescope.gitbooks.io/worldwide-telescope-web-control-script-reference/content/webcontrolobjects.html#wwtcontrol-loadtour-method

to load, play, and stop tours. We should add these to ``BaseWWTWidget`` (see ``center_on_coordinates`` for how to do this) and also add a corresponding handler in ``pywwt_web/wwt_json_api.js``

astrofrog:
Done in #28

PR #28, opened 2017-11-30
Added methods to load/control tours [issue #4]

ojustino:
I don't know if this is intentional, but `play_tour()` resumes playback immediately while `stop_tour()` slows the tour down to a stop over the course of a few seconds.

### DIFF
--- a/pywwt/core.py
+++ b/pywwt/core.py
@@ -58,6 +58,16 @@ class BaseWWTWidget(HasTraits):
 
     # TODO: need to add more methods here.
 
+    def load_tour(self,url):
+        # should this throw an error if url doesn't end in '.wwt'?
+        self._send_msg(event='load_tour',url=url)
+
+    def play_tour(self):
+        self._send_msg(event='play_tour')
+
+    def stop_tour(self):
+        self._send_msg(event='stop_tour')
+
     def center_on_coordinates(self, coord, fov, instant=True):
         coord_icrs = coord.icrs
         self._send_msg(event='center_on_coordinates',
--- a/pywwt/core.py
+++ b/pywwt/core.py
@@ -59,7 +59,6 @@ class BaseWWTWidget(HasTraits):
     # TODO: need to add more methods here.
 
     def load_tour(self,url):
-        # should this throw an error if url doesn't end in '.wwt'?
         """
         Load and begin playing a tour based on the URL to a .wtt file from 
         the WorldWideTelescope website.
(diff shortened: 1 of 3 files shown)


astrofrog:
Looks good! I'm not sure if that behavior is intentional - @philrosenfield, do you know?

Could you add some docstrings to the new methods? See:

https://github.com/WorldWideTelescope/pywwt/blob/master/pywwt/imagery.py#L8

for an example. We should follow the following guidelines for writing docstrings which means that parameters should be written as e.g.

```
Parameters
-----------
url : str
    The URL of the tour etc.
```

pywwt/core.py
```diff
@@ -58,6 +58,16 @@ def _send_msg(self, **kwargs):
 
     # TODO: need to add more methods here.
 
+    def load_tour(self,url):
+        # should this throw an error if url doesn't end in '.wwt'?
```
astrofrog: I don't think it makes sense to trust the extension only. But we should check what happens if the type of file is incorrect. Does the javascript print an error message?

philrosenfield:
I don't know if play_tour/stop_tour timing is intentional, but I'd assume it is. Is there a problem with it?

ojustino:
To Tom -- Do you want docstrings on every method in this pull request, or just for the tour-related ones?

To Phil -- I was assuming they would both act the same way, but upon further thought, that's more of an aesthetic expectation on my part than an actual problem.

astrofrog:
@ojustino - we need docstrings on all methods but you don't have to add them all in this PR if you'd rather not - up to you! (I just meant the tour ones for now)

pywwt/core.py
ojustino: Currently there's no message. Nothing happens in the widget, either.

ojustino:
I'll add docstrings for other methods in a separate pull request.

pywwt/core.py
astrofrog: Ok - having looked a bit at the docs I think we do want to check in here that the url ends in ``.wtt``. Could you add this?

pywwt/imagery.py
```diff
@@ -10,6 +10,11 @@ def get_imagery_layers(url):
     Get the list of available image layers that can be used as background
     or foreground based on the URL to a WTML (WorldWide Telescope image
     collection file).
+
+    Parameters
+    ----------
+    url : str
+        The URL of the chosen layer.
```
astrofrog: This should be *The URL of the image collection*

astrofrog:
Looks good, thanks!

Issue #25, opened 2017-11-29
Make some traitlets take astropy quantities


astrofrog:
I think the following properties should accept astropy quantities:

```
    location_altitude  = Float(0, help='Assigns altitude (in meters) for view location').tag(wwt='locationAltitude', sync=True)
    location_latitude  = Float(47.633, help='Assigns latitude for view location').tag(wwt='locationLat', sync=True)
    location_longitude = Float(122.133333, help='Assigns longitude for view location').tag(wwt='locationLng', sync=True)
```

I think the way to do that is to add a ``@validate`` for each of these (see e.g. the one for foreground_opacity) that converts the given values to the required units above and then returns the plain scalar in the right units.

astrofrog:
Done in #29

PR #29, opened 2017-12-01
Making local traits take astropy quants [issue #25]

ojustino:
1) I had to change the trait types from float to Any, but the unit check a) makes sure `proposal` is an astropy quantity, and b) raises an error if it's not the right type of quantity. Given that traitlets doesn't include data types with units, is this an acceptable workaround? 

2) Do you think it should still be possible for proposal to be a float and users have the option to choose which type they prefer? If it's a float, we could turn the unit check into an `elif` and first check if `proposal` is a float. 

3) What effect is location_altitude supposed to have in the widget? I see no discernible changes in the picture, even at Everest-level heights.

4. Is there a way to include all units of length for `location_altitude`? `find_equivalent_units()` was the best I found in astropy, but it leaves out km and other somewhat common units of height. Also, should imperial (or other) unit systems be included?

### DIFF
--- a/pywwt/core.py
+++ b/pywwt/core.py
@@ -107,15 +107,15 @@ class BaseWWTWidget(HasTraits):
             raise TraitError('location_altitude not in units of length')
    
     @validate('location_latitude')
-    def _validate_latiitude(self,proposal):
-        if proposal['value'].unit == u.degree:
+    def _validate_latitude(self,proposal):
+        if proposal['value'].unit.physical_type == 'angle':
             return proposal['value']
         else:
             raise TraitError('location_latitude not in degrees')
 
     @validate('location_longitude')
     def _validate_longitude(self,proposal):
-        if proposal['value'].unit == u.degree:
+        if proposal['value'].unit.physical_type == 'angle':
             return proposal['value']
         else:
             raise TraitError('location_longitude not in degrees')
--- a/pywwt/core.py
+++ b/pywwt/core.py
@@ -2,6 +2,7 @@ from traitlets import Bool, HasTraits, Float, Unicode, Any, observe, validate, T
 from astropy import units as u
 
 from .annotation import Circle, Poly, PolyLine
+from .quantity_trait import AstropyQuantity
 from .imagery import get_imagery_layers
 
 # The WWT web control API is described here:
@@ -96,9 +97,9 @@ class BaseWWTWidget(HasTraits):
                        instant=instant)
 
     local_horizon_mode = Bool(False, help='Whether the view should be that of a local latitude, longitude, and altitude').tag(wwt='localHorizonMode', sync=True)
-    location_altitude  = Any(0, help='Assigns altitude (in meters) for view location').tag(wwt='locationAltitude', sync=True)
-    location_latitude  = Any(47.633, help='Assigns latitude for view location').tag(wwt='locationLat', sync=True)
-    location_longitude = Any(122.133333, help='Assigns longitude for view location').tag(wwt='locationLng', sync=True)
+    location_altitude  = AstropyQuantity(0 * u.m, help='Assigns altitude (in meters) for view location').tag(wwt='locationAltitude', sync=True)
+    location_latitude  = AstropyQuantity(47.633 * u.deg, help='Assigns latitude for view location').tag(wwt='locationLat', sync=True)
+    location_longitude = AstropyQuantity(122.133333 * u.deg, help='Assigns longitude for view location').tag(wwt='locationLng', sync=True)
     
     @validate('location_altitude')
     def _validate_altitude(self,proposal):
--- a/pywwt/core.py
+++ b/pywwt/core.py
@@ -109,16 +109,16 @@ class BaseWWTWidget(HasTraits):
     @validate('location_latitude')
     def _validate_latitude(self,proposal):
         if proposal['value'].unit.physical_type == 'angle':
-            return proposal['value']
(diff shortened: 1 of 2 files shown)


astrofrog:
Thanks for working on this!

1. I think there is likely a way that we could define our own trait type that would be specific to astropy quantities. Maybe you could investigate the traitlets library a bit to see how one might define ``AstropyQuantity`` as a trait type that we could use rather than ``Any``? If this isn't easy, then in the validators, make sure you first check that ``isinstance(proposal['value'], Quantity)`` is True.

2. I think we should just strictly require quantities to avoid any ambiguity

3. I think the location matters for the position of the planets for example, but I'm not sure how much you have to change the values to see an effect.

4. The easiest way is to check ``value.unit.is_equivalent_to(u.m)`` or ``value.unit.physical_type == 'length'``

At the moment, if a user checks the value of e.g. ``location_longitude``, they will get a float back. I think we want to keep it as a quantity internally and then convert it only at the point where we send the message. To do this I think the easiest would be that in the validator you make sure you convert the values to meters and degrees but keep them as Quantity, then when sending the message in ``_on_trait_change`` you check if the value is a quantity and if so do ``.value`` before sending the message. Does this make sense?

ojustino:
I'm looking more into your suggestion for number 1. I think I correctly implemented your suggestion for saving the `.value` call until just before the message is sent.

ojustino:
I figured out how to make a custom trait in `quantity_trait.py` and got good results. Is the `hasattr()` check in the `validate()` method from that file good enough? I tried to keep it simple but still exclusive.

pywwt/core.py
```diff
@@ -26,11 +27,17 @@ def _on_trait_change(self, changed):
...
         # metadata attribute (which indicates the name of the corresponding WWT
         # setting).
-        wwt_name = self.trait_metadata(changed['name'], 'wwt')
+        wwt_name  = self.trait_metadata(changed['name'], 'wwt')
+        new_value = changed['new']
         if wwt_name is not None:
-            self._send_msg(event='setting_set',
-                           setting=wwt_name,
-                           value=changed['new'])
+            if hasattr(new_value, 'value') and hasattr(new_value, 'unit'):
```
astrofrog: I think it would be cleaner here to do ``if isinstance(new_value, Quantity):`` where ``Quantity`` is the Astropy Quantity class. Maybe then you could even do:

```python
if isinstance(new_value, Quantity):
    new_value = new_value.value
```

then just have ``_send_msg`` once below.

pywwt/quantity_trait.py
```diff
@@ -0,0 +1,12 @@
+from traitlets import TraitType
+from astropy import units as u
+
+class AstropyQuantity(TraitType):
+
+    default = 0 * u.deg
+    info_text = 'Custom trait to handle astropy quantities with units'
+
+    def validate(self, obj, value):
+        if hasattr(value, 'value') and hasattr(value, 'unit'):
```
astrofrog: As before, check ``isinstance(value, Quantity)`` instead of checking value and unit separately

pywwt/core.py
```diff
@@ -66,6 +66,32 @@ def center_on_coordinates(self, coord, fov, instant=True):
...
+    
+    @validate('location_altitude')
+    def _validate_altitude(self,proposal):
+        if proposal['value'].unit.physical_type == 'length':
+            return proposal['value'].to(u.meter)
+        else:
+            raise TraitError('location_altitude not in units of length')
+   
+    @validate('location_latitude')
+    def _validate_latiitude(self,proposal):
```
astrofrog: typo: latiitude -> latitude

pywwt/core.py
```diff
@@ -66,6 +66,32 @@ def center_on_coordinates(self, coord, fov, instant=True):
...
+        if proposal['value'].unit == u.degree:
+            return proposal['value']
```
astrofrog: You can do the same as for the altitude here: check the physical type is 'angle' then if so convert to degrees - it's unlikely anyone will give the lon/lat in other units, but you never know...

pywwt/quantity_trait.py
```diff
@@ -0,0 +1,12 @@
+from traitlets import TraitType
+from astropy import units as u
+
+class AstropyQuantity(TraitType):
+
+    default = 0 * u.deg
```
astrofrog: I don't think the default should have any specific units. Use ``0 * u.one`` to mean dimensionless (but still be an Astropy quantity)

pywwt/core.py
```diff
@@ -94,6 +94,32 @@ def center_on_coordinates(self, coord, fov, instant=True):
...
+    def _validate_altitude(self,proposal):
+        if proposal['value'].unit.physical_type == 'length':
+            return proposal['value'].to(u.meter)
+        else:
+            raise TraitError('location_altitude not in units of length')
+   
+    @validate('location_latitude')
+    def _validate_latitude(self,proposal):
+        if proposal['value'].unit.physical_type == 'angle':
+            return proposal['value']
```
astrofrog: Here make sure you include ``.to(u.degree)`` (and below too)

pywwt/core.py
```diff
@@ -94,6 +94,32 @@ def center_on_coordinates(self, coord, fov, instant=True):
...
+        else:
+            raise TraitError('location_latitude not in degrees')
```
astrofrog: Here maybe say 'not in angle units'

ojustino:
Apologies for all of the little things with this one.

PR #33, opened 2017-12-05
Added methods for returning RA and dec of view

ojustino:
I thought it would be a good idea to work on more methods based on WWTControl objects (foreground-background and tour-mth are branches that focused on already completed examples of these). Here I add get_ra() and get_dec(). Can you open and close the Qt widget and test them a few times? They seem to work half of the time and return `None` the other half of the time.

The two implemented methods that we haven't yet added either remove a selected annotation from the view or clear all annotations. Should removing a selected annotation be a method of the shape object (Circle/Poly/PolyLine) or of the widget (BaseWWTWidget)?


astrofrog:
@ojustino - thanks! I'll test this out.

Regarding removing annotations, I think this should be a method on the annotation itself for removing a single annotation, and a method on the base class of the widgets for removing *all* annotations.

pywwt/core.py
```diff
@@ -62,6 +62,24 @@ def _send_msg(self, **kwargs):
...
+        """
+        return self._send_msg(event='clear_annotations')
+    
+    def get_dec(self):
+        """
+        Return declination for the current view.
+        """
+        return self._send_msg(event='get_dec')
+
+    def get_ra(self):
```
astrofrog: Rather than have two separate methods for RA and Dec, it would be nice to have a single ``get_center`` method that returns an Astropy SkyCoord - could you change that?

ojustino:
I consolidated `get_ra()` and `get_dec()`. I added a test, but it passes whether I say `M42 == self.widget.get_center()` or `M42 != self.widget.get_center()`, which can't be the case. It works manually, but I don't think I wrote the test correctly.

ojustino:
I changed a few names -- polyline as we discussed, poly to polygon because I thought it was more clear, and `play_tour()` to `resume_tour()` since tours automatically start playing when loaded.

pywwt/tests/test_qt_widget.py
```diff
@@ -45,6 +45,11 @@ def test_methods(self, capsys):
         self.widget.center_on_coordinates(M42, fov=10 * u.deg)
         wait_and_check_output(1, capsys)
 
+    def test_coordinates(self, capsys):
+        self.widget.center_on_coordinates(M42, fov=10 * u.deg)
+        M42 == self.widget.get_center()
```
astrofrog: You need to add ``assert`` at the start of the line here to make sure the test fails if the value is False. Instead of checking equality (which won't work anymore) you could use the ``separation`` method on coordinates and check the separation is less than a certain threshold - here's an example of how to use it:

```
In [6]: SkyCoord.from_name('M32').separation(SkyCoord.from_name('M42')).arcsec
Out[6]: 291511.6229661299
```

ojustino:
OK, the test works now. The actual value of the separation in arcseconds is 1.5279869286995402e-10.

ojustino:
I made changes based on the question I asked on Slack a couple of days ago. I did end up having to use Python to get new circles to center in the current view instead of (0,0) -- I tried a few ways of doing the math in JavaScript but they all ended up with incorrect values.

astrofrog:
Looks good! I'll merge now :)

PR #37, opened 2017-12-05
Implement method to render to file and added image test

### LINKED ISSUE #35: Set up pixel-by-pixel image test
I'm not sure how feasible this is going to be in terms of how the rendering might be different on different machines, but worth the try.

### LINKED ISSUE #36: Add method to render content of Qt widget to PNG/JPG
I think this would be most easily done on the Qt widget for now

astrofrog:
This adds a method to render the view from the Qt widget to a file, and adds a test that makes sure that the various settings are working.

@ojustino - just for info, I fixed a couple of things that came up during the testing - for example the default annotation values were not being set, which caused some weird issues with skyRelative - but this is all fixed now.

Fixes #35 
Fixes #36


astrofrog:
@ojustino - just for information, I'm merging this. If you take a look at 'Files changed' you'll see that there are now reference images and a test that makes sure the output stays the same.

Issue #17, opened 2017-11-17
Add validation to traits/settings


astrofrog:
Once https://github.com/astrofrog/pywwt-web/issues/1 is done, we should add some further validation to the traits. For colors, we could use e.g. matplotlib to understand different ways of specifying the color and converting it to hex. For other kinds of values we should test e.g numerical value ranges. See http://traitlets.readthedocs.io/en/stable/using_traitlets.html#validation for information on adding validation to traits.

astrofrog:
Done in #38

PR #38, opened 2017-12-05
Added validators for circle radius and all linewidths

ojustino:
Color validators and comment removal coming next.

### DIFF
--- a/pywwt/annotation.py
+++ b/pywwt/annotation.py
@@ -1,6 +1,7 @@
 import uuid
-from traitlets import HasTraits, TraitError, validate
+from traitlets import Any, HasTraits, TraitError, validate
 from astropy import units as u
+from matplotlib import colors
 
 from .traits import Bool, Float, Unicode, AstropyQuantity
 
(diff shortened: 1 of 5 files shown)


ojustino:
`colors.to_hex()` accepts tuples of length 3 and 4, with the 4th entry representing opacity. WWT annotation colors don't have opacities, so should an error be thrown if a user tries something like `circle.line_color = (.7,.33,.1,.6)`? At the moment I believe the color specified by the tuple's first three entries is successfully applied as if the fourth entry didn't exist.

Also, should the type in the help for the color traits be "(:class:`str`) or (:class:`tuple`)?"

astrofrog:
@ojustino - hmm good question. Actually annotations do have an opacity property, so if the color passed to matplotlib has an opacity, maybe we could automatically set the opacity property? (but maybe for now, fine to raise an error)

astrofrog:
Fixes https://github.com/WorldWideTelescope/pywwt/issues/17

(this will cause that issue to get closed when this is merged)

ojustino:
Once we convert to hex, there's no way to recover the opacity, so the _send_msg for opacity would have to take place before the `return` in the validate functions. Should this occur in the validate functions or is there another decorator I can do it in?

ojustino:
I attempted to add opacity to the "fill_color" option for circles. It works, and if you think the execution is good I can do the same for the other shapes and constellation traits.

pywwt/annotation.py
```diff
@@ -52,11 +53,42 @@ class Circle(Annotation):
...
+
+    @validate('fill_color')
+    def _validate_fillcolor(self, proposal):
+        if isinstance(proposal['value'], str):
+            return colors.to_hex(proposal['value'])
+        elif isinstance(proposal['value'], tuple):
+            if len(proposal['value']) == 3:
+                return colors.to_hex(proposal['value'])
+            if len(proposal['value']) == 4:
+                self.parent._send_msg(event='annotation_set',
```
astrofrog: Instead of doing this, I think you should set ``self.opacity = ...`` otherwise ``circle.opacity`` will be out of sync with the actual opacity in WWT.

astrofrog:
In addition to the comment above, add Matplotlib in the ``.travis.yml`` file on the line ``CONDA_DEPENDENCIES`` and also add ``matplotlib`` to ``install_requires`` in ``setup.py``

Issue #43, opened 2017-12-07
Make it possible to specify the polygon/line points during initialization


astrofrog:
When initializing a polygon or line, it would be nice to be able to specify the points, e.g.

```python
circle = wwt.add_polygon(points=SkyCoord(...))
```

this could internally just call ``add_point``.

astrofrog:
Done in #46

Issue #42, opened 2017-12-07
Support vector SkyCoords for add_point


astrofrog:
The SkyCoord class allows vectors of coordinates to be represented:

```
In [1]: from astropy.coordinates import SkyCoord

In [2]: c1 = SkyCoord.from_name('M31')

In [3]: c2 = SkyCoord([1, 2, 3], [3, 4, 5], unit='deg')

In [4]: c1.isscalar
Out[4]: True

In [5]: c2.isscalar
Out[5]: False

In [6]: c2.icrs.ra
Out[6]: <Longitude [ 1., 2., 3.] deg>

In [7]: c2.icrs.ra.degree
Out[7]: array([ 1.,  2.,  3.])
```

We should make ``add_point`` for polygons and lines accept vector SkyCoords so that many points can be added in one go.

PR #46, opened 2017-12-10
Enhancements

ojustino:


ojustino:
I realize there's an issue with the fix for #42 but I'm working on it along with #43.

pywwt/core.py
```diff
@@ -197,10 +198,16 @@ def create_circle(self):
                        dec=self.get_center().dec.deg)
         return circle
 
-    def add_polygon(self):
+    def add_polygon(self,points=None):
```
astrofrog: Here maybe you can also add ``**kwargs`` and pass that to ``Polygon`` below so that the annotation traits can also be set on initialization?

pywwt/qt_widget.py
```diff
@@ -127,7 +127,7 @@ def _run_js(self, js):
 
 class WWTQtWidget(BaseWWTWidget):
 
-    def __init__(self, block_until_ready=False):
+    def __init__(self, block_until_ready=False, size=None, **kwargs):
```
astrofrog: What is the ``size=None`` for?

ojustino: I was following the code from the issue, which included `size` as an argument. I don't see it anywhere else so I'll take it out.

ojustino:
Since `**kwargs` allows for any keyword argument, do you think it'd be a good idea to use `[annotation].trait_names()` to warn users if they provide one that is not part of that annotation's class?

astrofrog:
@ojustino - this will need to be rebased now that I merged your other pull request. Yes maybe checking that the kwargs are in ``trait_names`` would be sensible.

pywwt/annotation.py
```diff
@@ -27,19 +27,20 @@ class Annotation(HasTraits):
...
         self.parent._send_msg(event='annotation_create', id=self.id, shape=self.shape)
+        super(Annotation, self).__init__(**kwargs)
 
     def _on_trait_change(self, changed):
         # This method gets called anytime a trait gets changed. Since this class
         # gets inherited by the Jupyter widgets class which adds some traits of
         # its own, we only want to react to changes in traits that have the wwt
         # metadata attribute (which indicates the name of the corresponding WWT
         # setting).
+        print('_an_otc',changed,sep=' ')
```
astrofrog: This should be removed

pywwt/annotation.py
```diff
@@ -94,17 +95,23 @@ def _on_trait_change(self, changed):
 class Polygon(Annotation):
 
     shape = 'polygon'
-
+    
     fill = Bool(False, help='Whether or not the polygon should be filled').tag(wwt='fill', sync=True)
-    fill_color = Unicode('white', help='Assigns fill color for the polygon').tag(wwt='fillColor', sync=True)
+    fill_color = Unicode('blue', help='Assigns fill color for the polygon').tag(wwt='fillColor', sync=True)
```
astrofrog: I think this should be changed back to white

ojustino:
I wanted to let you see the results of the rebase. I'm working on the kwargs check, so please wait to merge if the branch is otherwise OK.

astrofrog:
@ojustino - this looks good. I think the reason the tests are failing is that i think the circle radius should be in degrees not arcseconds.

ojustino:
I changed the units, so we'll see how the checks go. I added the kwargs restriction to the Annotation class, too, so everything I planned to implement on this branch is now present.

astrofrog:
This looks great! You can ignore the AppVeyor CI failure for now, I'm working on that in a separate pull request.

PR #51, opened 2017-12-12
Method to load astropy tables [issue #5]

ojustino:
I attempted to solve the issue above. The function takes a column of SkyCoord entries from an astropy table and generates circles one by one. Currently, they all share a name, which isn't optimal, but I haven't yet found the way to make them all distinct annotations.

### DIFF
--- a/pywwt/annotation.py
+++ b/pywwt/annotation.py
@@ -100,7 +100,7 @@ class Circle(Annotation):
         elif proposal['value'].unit.is_equivalent(u.degree):
             return proposal['value'].to(u.degree)
         else:
-            raise TraitError('radius must be in pixel or arcsec equivalent unit')
+            raise TraitError('radius must be in pixel or degree equivalent unit')
 
     def set_center(self, coord):
         coord_icrs = coord.icrs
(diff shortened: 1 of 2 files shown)


astrofrog:
@ojustino - I think we are going to want to have a different approach for plotting multiple circles - I think we should make it possible for the user to get back an object to control the properties of the circles (like a ``CircleCollection``). We shouldn't put too much work into this however because I think the WWT developers are working on adding the 'layer' API to the web client, which is an efficient way to deal with many markers.

For now, if you are interested in trying to do the CircleCollection, I'd suggest making a new class called ``CircleCollection`` which basically contains references to multiple circles. Then if the user does:

```python
collection = wwt.add_circle_collection(skycoords)
```

they get a ``CircleCollection`` back. Doing:

```
collection.color = 'red'
```

would then internally change the color of all the circles. We could also implement ``__getitem__`` on the class so that you could do e.g.:

```
collection[0].color = 'green'
```

to modify just one of the circles. Do you have a feel for how to do this, or would you like me to show some example code for how to get started?

ojustino:
I'm trying to start working from this, but it would be nice to see some more example code when you get the chance (with the release going on).

PR #52, opened 2017-12-13
Implement Color trait

astrofrog:
@ojustino - could you review this?


astrofrog:
I just realized that we should only set the opacity if the color is the fill color - I need to make some tweaks for that

ojustino:
This looks good to me. I also realized that not all color attributes have opacity, and we agree on the traits for which that is or is not the case.

PR #54, opened 2017-12-16
CircleCollection

ojustino:
@astrofrog - I think I have a working start to implementing the `CircleCollection` class. Is this close to what you had in mind? (I renamed the branch that I originally tried the astropy tables solution on; I didn't mean to delete the old PR.) 

One issue I've seen so far with this attempt is that deleting individual circle instances isn't synced with deleting their references from `self.collection` and `self.points`, and I haven't yet figured out how to fix that. 

I could include `__delitem__` in the class definition, but I think it's more natural to delete a circle from `CircleCollection` "collect" via `collect[2].remove()` rather than `del collect[2]` given the syntax we created for the other annotations. Maybe I could have a decorator that watches for calls to `collect[x].remove()` and deletes item `x` from `self.collection` and `self.points`, but I'm not sure how since `remove()` belongs to the `Circle` class.

**EDIT: Upon further review, the references to circles and other annotations aren't deleted when you call `circle.remove()`; the shape is just removed from the viewer. `change_all()` here still works even after deleting circles, so does this mean that `CircleCollection`'s current behavior is OK?


astrofrog:
> One issue I've seen so far with this attempt is that deleting individual circle instances isn't synced with deleting their references from self.collection and self.points, and I haven't yet figured out how to fix that.

One way around this is to simply not implement ``__getitem__`` and instead always require operations to be done on all circles.

This is close to what I envisaged - maybe you could implement ``__setattr__`` to provide a more convenient interface to ``change_all``?

ojustino:
Would I be implementing `__setattr__` on `CircleCollection`? Right now I'm having trouble since `CircleCollection` has its own internal instance variables and the ones that we would like to change are actually from the `Circle` class.

astrofrog:
@ojustino - yes, I meant that if you implement ``__setattr__`` on CircleCollection then if someone tries to change e.g. CircleCollection.radius, it would then internally change all the Circle.radius. Does this make sense? If not just let me know and I can provide some example code.

ojustino:
I currently have this:

```
def __setattr__(self, name, value):
        if name in [i for i in dir(Circle) if not (i.startswith('__') or i.startswith('_') or i == 'parent')]:
            for point in self.collection:
                point.__dict__['_trait_values'][name] = value
                # point.name = value
        else:
            self.__dict__[name] = value
```
Given the example: 
```
In [8]: col.collection.line_width = 3 * u.pixel
```
...these are my problems.

1) I have the `if` condition to try and differentiate instance variables meant for the `Circle` objects from those meant for the `CircleCollection`, but I feel like there's a better way to check.

2) Once the if condition is met, I would like to have `point.name = value`, but `name` is being taken as a string (`repr(name) == 'line_width'`), which means I can't use it to access object attributes that way. 

3. Using `point`'s `__dict__` changes the value of both `point.__dict__['_trait_values'][name]` and `point.name` to 3.0 pix like I want, but the circles in the viewer don't change.

Currently, the only way I can change the circles' appearance is one at a time in the interpreter, like so: `In [9]: col.collection[0].line_width = 3 * u.pixel`.

ojustino:
Does the list comprehension work as a solution to 1) above? I also added a `__getattr__` method that returns a list in case not all circles have the same value for a particular attribute.

pywwt/annotation.py
```diff
@@ -230,3 +231,44 @@ def _on_trait_change(self, changed):
...
 
         super(Line, self)._on_trait_change(changed)
+
+class CircleCollection():
+    
+    def __init__(self, parent, points, **kwargs): # save args, make first circles
+        if len(points) <= 1e4: # subject to change
+            self.points = points
+        else:
+            raise IndexError('too many points') # correct error?
```
astrofrog: Maybe change this error to 'For performance reasons, only 10000 annotations can be added in one go at this time'

pywwt/annotation.py
```diff
@@ -230,3 +231,44 @@ def _on_trait_change(self, changed):
...
+        self.parent = parent
+        self.collection = []
+        self.gen_circles(self.points, **kwargs)
+        
+    def __setattr__(self, name, value):
+        if name in [key for key in Circle.__dict__ if not key.startswith('_')]:
+            for elem in self.collection:
+                setattr(elem, name, value)
+                #print(elem.__dict__['_trait_values'][name])
```
astrofrog: Remove debug print statement

pywwt/annotation.py
```diff
@@ -230,3 +231,44 @@ def _on_trait_change(self, changed):
...
+            values = []
+            for elem in self.collection:
+                attr = getattr(elem, name)
+                if attr not in values:
+                    values.append(attr)
+            return values
+        else:
+            raise AttributeError('CircleCollection object has no attribute '+repr(name))
+        
+    def gen_circles(self, points, **kwargs):
```
astrofrog: Since I don't think ``gen_circles`` is meant to be used, maybe make it private, i.e. ``_gen_circles``

astrofrog:
@ojustino - there is a way to automate it a little or at least reduce boilerplate code, but I propose that we merge this as-is (once you remove the comments) and I'll open an issue describing how I think we can simplify it that we can work on as part of the next contract (since it's not necessarily trivial)

PR #55, opened 2017-12-18
Tidy up docs for release

astrofrog:
@ojustino - just FYI, there is a change here that might break some scripts you have which is that ``WWTQtWidget`` has been renamed to ``WWTQtClient``.

Another change I made is to change the Python examples to use ``>>>`` instead of IPython prompts - using ``>>>`` makes it easier to avoid having to update the prompt numbers, and there's also a plugin I'll add shortly which makes it easier to copy/paste code from ``>>>`` blocks (ignoring the ``>>>``)


ojustino:
@astrofrog Thanks for letting me know. I noticed some typos in these edits; should I list them here or open a new PR that fixes them?

astrofrog:
@ojustino - thanks for spotting the typos - could you open a PR to fix them? Thanks!

PR #58, opened 2017-12-18
Docstrings

ojustino:
I tried to cover everything on [this page](http://pywwt.readthedocs.io/en/latest/api.html) and all of the subsequent links besides Windows.

### DIFF
--- a/pywwt/annotation.py
+++ b/pywwt/annotation.py
@@ -59,15 +59,14 @@ class Annotation(HasTraits):
 
 class Circle(Annotation):
     """
-    Used to render a circle on screen (via pywwt.BaseWWTWidget.add_circle()) 
-    and edit its properties after intialization. Inherits from Annotation
-    class.
+    Used to render a circle on screen
+    (via :meth:`pywwt.BaseWWTWidget.add_circle`) and edit its properties
+    after intialization.
     """
 
     shape = 'circle'
     """
-    The shape to be sent to the Annotation class for object creation 
-    (:class:`str`)
+    The name of the shape (:class:`str`)
     """
 
     fill = Bool(False, help='Whether or not the circle should be filled (:class:`bool`)').tag(wwt='fill')
@@ -94,7 +93,7 @@ class Circle(Annotation):
 
     def set_center(self, coord):
         """
-        Set the center coordinates of a circle object.
+        Set the center coordinates of the circle object.
 
         Parameters
         ----------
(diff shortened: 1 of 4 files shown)


pywwt/annotation.py
```diff
@@ -52,8 +58,17 @@ def _on_trait_change(self, changed):
 
 
 class Circle(Annotation):
+    """
+    Used to render a circle on screen (via pywwt.BaseWWTWidget.add_circle()) 
+    and edit its properties after intialization. Inherits from Annotation
+    class.
```
astrofrog: You don't need to say 'Inherits from ...'

Also for ``pywwt.BaseWWTWidget.add_circle()``, write:

```
:meth:`pywwt.BaseWWTWidget.add_circle`
```

instead and that will auto-link to the API docs

pywwt/annotation.py
```diff
@@ -52,8 +58,17 @@ def _on_trait_change(self, changed):
...
+    """
 
     shape = 'circle'
+    """
+    The shape to be sent to the Annotation class for object creation 
+    (:class:`str`)
+    """
```
astrofrog: I would just say 'The name of the shape' to keep it simple

pywwt/annotation.py
```diff
@@ -78,12 +93,23 @@ def _validate_radius(self, proposal):
             raise TraitError('radius must be in pixel or arcsec equivalent unit')
 
     def set_center(self, coord):
+        """
+        Set the center coordinates of a circle object.
```
astrofrog: of a circle object -> of the circle object

pywwt/annotation.py
astrofrog: Same comment for the other shapes

pywwt/core.py
```diff
@@ -117,16 +138,34 @@ def center_on_coordinates(self, coord, fov=60*u.deg, instant=True):
...
+
+        Parameters
+        ----------
+        dt : `~datetime.datetime` or `~astropy.time.core.Time`
+            The current time, either as a `datetime` object or an astropy
+            `Time`.
+        """
+        if isinstance(dt, t.core.Time):
+            dt = dt.datetime
+            self._send_msg(event='set_datetime',
```
astrofrog: Just un-indent the self._send_msg and then remove the elif statement below. That is, you can just do:

```
if isinstance(dt, t.core.Time):
    dt = dt.datetime
self._send_msg(...)
```

pywwt/jupyter.py
```diff
@@ -22,6 +22,10 @@
 
 @widgets.register
 class WWTJupyterWidget(widgets.DOMWidget, BaseWWTWidget):
+    """
+    Defines the Python part of the WWT Jupyter widget. Inherits from
+    BaseWWTWidget class.
```
astrofrog: Remove the 'Inherits' bit

PR #63, opened 2018-01-24
Proper support for Jupyter lab

astrofrog:
This is just an experimental branch as I'm at a Jupyter widgets workshop this week.

Things to do/figure out:

* [x] Preserve compatibility with notebook, might need to serve separate static files from notebook version
* [ ] Whether using a server extension to serve the html and js file is going to be problematic for e.g. binder or Azure notebook
* [ ] How to automatically install the server extension
* [ ] How to improve performance of WWT when split into a side panel
* [ ] Better resize canvas based on available space so it works in tabs/side panels

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 0.4.0 (unreleased)
 ------------------
 
-- No changes yet.
+- Fix compatibility with Jupyter Lab. [#63]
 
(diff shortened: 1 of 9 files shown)


astrofrog:
Some of the TODOs are going to be hard to do without actually merging this and testing it out on binder, so once this is otherwise ready, I'll merge this.

PR #64, opened 2018-01-24
Added GUI controls for layers

astrofrog:
This adds GUI controls for setting the foreground, background, and foreground opacity. These controls can be shown by doing ``wwt.layer_controls``. I want to add a mention of this to the docs. We could also decide to show these by default with the WWT widget, but some users might not want it so it could be an option.

**Preview:**

screen shot 2018-01-24 at 16 12 09

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,8 @@
 
 - Fix compatibility with Jupyter Lab. [#63]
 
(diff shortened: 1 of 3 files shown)


astrofrog:
I'll merge this once tests pass then will add docs later once the functionality has converged

PR #70, opened 2018-02-22
Adding to/polishing docs

ojustino:

PR #71, opened 2018-02-25
Attempted solution for tab completing layers

ojustino:
Here's my method to a) allow users to tab complete layer names, and b) automatically sort new layers by wavelength band.

I combined your proof-of-concept with regular expressions to look for key words in the strings that comprise the layer names ('UV', 'microwave', etc.). This way, if a user decides to upload more layers, they'll automatically be sorted.

I also use regular expressions to shorten the layer names and specifically change '2MASS' to 'twomass'. I also read about a package that converts integers to their string representations (e.g. 6 to 'six'); let me know if you want me to integrate into `layers.py` as well.

Then, I add one line to `core.py`, and it allows for layers to be changed like so:
`wwt.foreground = wwt.imagery.x.rass`.

I'm not sure how to check if properties show up in the docs, or if you would want these to do so in the first place.

### DIFF
--- a/pywwt/layers.py
+++ b/pywwt/layers.py
@@ -21,47 +21,39 @@ class ImageryLayers():
         # Helps turn the list of layer names used to initialize the class
         # (og_list) into a dict.
         for layer in og_list:
-            ind = 0
 
             if re.search(r'(?i)gamma',layer) is not None:
                 self._add2dict(self._layers, layer, 'gamma')
                 continue # automatically advance to next iteration
 
-            ind += 1
             if re.search(r'(?i)x(-|\s)?ray',layer) is not None:
                 self._add2dict(self._layers, layer, 'x')
                 continue
 
-            ind += 1
             if (re.search(r'(?i)ultra(-|\s)?violet',layer) is not None or
                 re.search(r'(?i)[^\d\w]+uv|uv[^\d\w]+',layer) is not None):
                 self._add2dict(self._layers, layer, 'uv')
                 continue
 
-            ind += 1
             if (re.search(r'(?i)optical',layer) is not None or
                 re.search(r'(?i)visible',layer) is not None):
                 self._add2dict(self._layers, layer, 'visible')
                 continue
 
-            ind += 1    
             if (re.search(r'(?i)infrared',layer) is not None or
                 re.search(r'(?i)[^\d\w]+ir|ir[^\d\w]+',layer) is not None):
                 self._add2dict(self._layers, layer, 'ir')
                 continue
 
-            ind += 1
             if (re.search(r'(?i)microwave',layer) is not None or
                 re.search(r'(?i)[^\d\w]+cmb|cmb[^\d\w]+',layer) is not None):
                 self._add2dict(self._layers, layer, 'micro')
                 continue
 
-            ind += 1
             if re.search(r'(?i)radio',layer) is not None:
                 self._add2dict(self._layers, layer, 'radio')
                 continue
 
-            ind += 1
             self._add2dict(self._layers, layer, 'other')
 
     def _add2dict(self, diction, full_layer, bandpass):
--- a/pywwt/layers.py
+++ b/pywwt/layers.py
@@ -24,45 +24,45 @@ class ImageryLayers():
             ind = 0
 
             if re.search(r'(?i)gamma',layer) is not None:
-                self._add2dict(self._layers, layer, self._spectrum[ind])
+                self._add2dict(self._layers, layer, 'gamma')
                 continue # automatically advance to next iteration
 
             ind += 1
             if re.search(r'(?i)x(-|\s)?ray',layer) is not None:
-                self._add2dict(self._layers, layer, self._spectrum[ind])
+                self._add2dict(self._layers, layer, 'x')
                 continue
 
             ind += 1
             if (re.search(r'(?i)ultra(-|\s)?violet',layer) is not None or
                 re.search(r'(?i)[^\d\w]+uv|uv[^\d\w]+',layer) is not None):
-                self._add2dict(self._layers, layer, self._spectrum[ind])
+                self._add2dict(self._layers, layer, 'uv')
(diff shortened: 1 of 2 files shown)


pywwt/layers.py
```diff
@@ -0,0 +1,100 @@
...
+    def __init__(self, layer_list):
+        self._layers = {}
+        self._spectrum = ['gamma', 'x', 'uv', 'visible', 'ir', 'micro', 'radio', 'other']
+        for band in self._spectrum:
+            self._layers[band] = {}
+        
+        self._list2dict(layer_list)
+
+    def _list2dict(self, og_list):
+        for layer in og_list:
```
astrofrog: I think it would be good to include a comment here about what this method does and what ``og_list`` is expected to be.

pywwt/layers.py
```diff
@@ -0,0 +1,100 @@
...
+            ind = 0
+
+            if(re.search(r'(?i)gamma',layer) != None):
```
astrofrog: The parentheses around the whole if expression are not needed - also for ``None`` one normally writes ``is not None`` (same comment for if statements below)

pywwt/layers.py
```diff
@@ -0,0 +1,100 @@
...
+                self._add2dict(self._layers, layer, ind)
+                continue # automatically advance to next iteration
+
+            ind += 1
```
astrofrog: I am assuming that ``ind`` gives the position in the spectrum list of bandpasses? If so I wonder whether it would be cleaner if ``add2dict`` simply took the name of the bandpass instead of these indices?

pywwt/layers.py
```diff
@@ -0,0 +1,100 @@
...
+        abbr = self._shorten(full_layer)
+
+        # check if abbr already exists in diction
+        j = 2
+        while(abbr in diction[self._spectrum[ind]]):
+            if(str(j-1) in abbr):
+                abbr = abbr[:-1]
+
+            abbr += str(j)
+            j += 1
```
astrofrog: I think the logic can be simplified here, something like:

```python
short =  self._shorten(full_layer)
suffix = ''
while short + str(suffix) in diction[self._spectrum[ind]]:
    suffix = suffix + 1 if suffix else 1
```

ojustino:
I added your simplifications as well as some comments that should make the code more clear. Some other questions:

1) What do you think about using a package ([num2words](https://pypi.python.org/pypi/num2words/0.5.6)) to generalize the string to integer translation process in `_shorten()`? At the moment, the code only addresses 2mass, though names with other numbers could conceivably show up once we can import more layers.

2) What was the 'thumbnail' key in `_add2dict()` meant to do? If they're all None, does it make sense to just leave it out?

astrofrog:
@ojustino - I think we should avoid adding a dependency on num2words, and in any case I don't think we want to translate e.g. 1999 to onethousandninehundredandninetynine ;) I think maybe just hard-code the translation for 0-9 and then beyond that we can add e.g. an underscore in front?

For ``_add2dict``, i think @philrosenfield wrote the original version.

ojustino:
OK, I went the hard-coding route and added a check for leading underscores and non-letter/number characters. We can wait to see what Phil says about 'thumbnail', but is everything else in order?

pywwt/layers.py
```diff
@@ -70,26 +73,34 @@ def _add2dict(self, diction, full_layer, bandpass):
...
-                short += str(suffix)
             else:
                 suffix = 1
-                short += str(suffix)
+                
+            short += str(suffix)
 
         diction[bandpass][short] = {}
         diction[bandpass][short]['full_name'] = full_layer
         diction[bandpass][short]['thumbnail'] = None
```
philrosenfield: I think 'thumbnail' can be done away with unless there is a reason to keep all passable options to the JS API.  I can imagine this may be useful later on if:
- someone was making a pywwt GUI and wanted to use a thumbnail image as is used on the wwt web client. 
- someone wanted to send a custom layer from pywwt to the WWT webclient using this class

However, both are beyond the scope of this feature.

pywwt/layers.py
```diff
@@ -0,0 +1,125 @@
...
+        for layer in og_list:
+            ind = 0
+
+            if re.search(r'(?i)gamma',layer) is not None:
+                self._add2dict(self._layers, layer, self._spectrum[ind])
+                continue # automatically advance to next iteration
+
+            ind += 1
+            if re.search(r'(?i)x(-|\s)?ray',layer) is not None:
+                self._add2dict(self._layers, layer, self._spectrum[ind])
```
astrofrog: Just to check, why not just put ``'x'`` instead of ``self._spectrum[ind]`` here? I think it would be more readable if you just hard-code the bandpasses instead of relying on ``ind``.

pywwt/layers.py
ojustino: OK, thanks for letting me know. We can keep it mind in case we widen this feature's scope in the future.

pywwt/layers.py
ojustino: I was prioritizing portability over readability, but I can hard-code the bandpass names in the next commit.

pywwt/layers.py
```diff
@@ -0,0 +1,124 @@
...
+                re.search(r'(?i)[^\d\w]+cmb|cmb[^\d\w]+',layer) is not None):
+                self._add2dict(self._layers, layer, 'micro')
+                continue
+
+            ind += 1
+            if re.search(r'(?i)radio',layer) is not None:
+                self._add2dict(self._layers, layer, 'radio')
+                continue
+
+            ind += 1
```
astrofrog: I think you can also remove all the ``ind``s?

astrofrog:
This looks good! Before we merge, could you add some documentation about this?

ojustino:
I did; it's in #70 since I already had another docs-based PR open.

astrofrog:
Ok perfect, I'll go ahead and merge this, thanks!

Issue #69, opened 2018-02-22
Add support for solar system mode


astrofrog:
I think the web client does have support for the solar system mode - if so, we should add support for it in pywwt.

astrofrog:
Yep! :)

PR #74, opened 2018-03-20
Solar system mode and associated methods

ojustino:
First attempt, in reference to issue #69


pywwt/core.py
```diff
@@ -191,16 +191,36 @@ def _validate_longitude(self, proposal):
...
+        Parameters
+        ----------
+        mode : `str`
+            The desired mode. (default: 'sky')
+        """
+        available = ['sky', 'sun', 'mercury', 'venus', 'earth', 'moon', 'mars',
+                     'jupiter', 'saturn', 'uranus', 'neptune', 'pluto']
+        ss_mode = '3D Solar System View'
+        
+        if mode in available:
```
philrosenfield: users shouldn't need to worry about case sensitivity.  Perhaps `if mode.lower() in available` or ValueError prints available + 'solar_system'

pywwt/core.py
```diff
@@ -191,16 +191,36 @@ def _validate_longitude(self, proposal):
...
+            self._send_msg(event='set_viewer_mode', mode=mode)
+        elif mode == 'solar_system':
```
philrosenfield: This is a personal issue with WWT, but the "3D solar system mode" has an infinite zoom, as well as SDSS Cosmos data. I think this might be a chance to call this mode '3D' or '3d universe' or something you all think is better than 'solar_system'

pywwt/core.py
astrofrog: I agree!

pywwt/core.py
ojustino: OK, I'm open to suggestions on the name.

ojustino:
Looking through the source code that Phil linked to helped me find a lot more methods, and I added as many as I could in the latest commit. Results of note:

1. I made comments in `wwt.html` to indicate which solar system mode related methods are working and which are not. By the way, any better names for solar system mode? I agree with Phil that it does more than show the solar system and its name should reflect that.

2. I got planet tracking to work through the method `wwt.track_object()`. It takes a string that will point to the integer used as an argument for the actual JS function. I went to Earth and it was rotating, which was cool. Individual views in planet mode are also available for the Galilean satellites.

3. Panorama views are now available, although I don't know how to switch to the other ones. Are they hidden like the imagery layers for minor planets?

4. I found a cause of the `libpng warning: gAMA: gamma value out of range` errors that occur for me when I launch in Qt -- the images that load when the new `wwt.constellation_pictures` attribute is toggled on. Is there way not to load these pictures on startup until we figure out how to fix the gamma value? (I also sometimes get `libpng` errors when I increase the scale of the planets to 100 in solar system mode.)

5. I also added `wwt.constellation_labels`, which should label the constellations, but I get an error saying that the XML requests necessary to run this method are not allowed. A similar thing happens with some other labeling attributes I added for the (also newly added) ecliptic and galactic grids.

6. Crosshairs are no longer working, so I couldn't test the method I added that allows users to change their color.

ojustino:
Looking through the source code that Phil linked to helped me find a lot more methods, and I added as many as I could in the latest commit. Results of note:

1. I made comments in `wwt.html` to indicate which solar system mode related methods are working and which are not. By the way, any better names for solar system mode? I agree with Phil that it does more than show the solar system and its name should reflect that.

2. I got planet tracking to work through the method `wwt.track_object()`. It takes a string that will point to the integer used as an argument for the actual JS function. I went to Earth and it was rotating, which was cool. Individual views in planet mode are also available for the Galilean satellites.

3. Panorama views are now available, although I don't know how to switch to the other ones besides the default Mars rover. Are these imagery layers hidden like the ones for minor planets, too?

4. I found a cause of the `libpng warning: gAMA: gamma value out of range` errors that occur for me when I launch in Qt -- the images that load when the new `wwt.constellation_pictures` attribute is toggled on. Is there way not to load these pictures on startup until we figure out how to fix the gamma value? (I also sometimes get `libpng` errors when I increase the scale of the planets to 100 in solar system mode.)

5. I also added `wwt.constellation_labels`, which should label the constellations, but I get an error saying that the XML requests necessary to run this method are not allowed. A similar thing happens with some other labeling attributes I added for the (also newly added) ecliptic and galactic grids.

6. Crosshairs are no longer working, so I couldn't test the method I added that allows users to change their color.

astrofrog:
Here are some answers from my side:

1. What about keeping 'solar system', but then having e.g. 'milky way' and 'universe' which are the same but start off more zoomed out?

2. Very cool :)

3. Maybe we can defer this to a separate PR so this one can be wrapped up first?

4. We need to find a way to try and get the actual image files to see if there are issues with them...

5. Sounds similar to the issue I was having with loadFITS. Let's see what they say about that.

pywwt/core.py
```diff
@@ -182,28 +193,59 @@ def _validate_longitude(self, proposal):
...
+                    'saturn': 5, 'uranus': 6, 'neptune': 7, 'pluto': 8,
+                    'moon': 9, 'io': 10, 'europa': 11, 'ganymede': 12,
+                    'callisto': 13, 'ioshadow': 14, 'europashadow': 15,
+                    'ganymedeshadow': 16, 'callistoshadow': 17,
+                    'suneclipsed': 18, 'earth': 19}
+
+        if obj in mappings:
+            self._send_msg(event='track_object', code=mappings[obj])
 
     def set_view(self, mode):
```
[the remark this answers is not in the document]
ojustino: The SDK doesn't take FOV here. I'm thinking the best solution is just to use the new `reset_view` function after you've switched modes in order to get back to a more normal FOV.

ojustino:
The main thing I do in this commit is give `BaseWWTWidget` a class variable that keeps track of the current mode (`current_mode`) and add a function (`reset_view()`) that resets the view based on `current_mode`. 

I also divided solar system mode into "solar_system," "milky_way," and "universe" modes and hard-coded default coordinates and FOVs for every mode except panorama, which I'll try to address in a future PR as per Tom's request.

There's no change in terms of which solar system methods work or with the other methods that were broken or causing issues, though I guess those are out of our hands at the moment.

pywwt/static/wwt.html
```diff
@@ -49,19 +49,41 @@
...
-        wwt.settings.set_constellationFigureColor('#ff0000')
         wwt.settings.set_showConstellationFigures(false)
         wwt.settings.set_showConstellationSelection(false)
+        //wwt.settings.set_showConstellationPictures(false)//libpng gamma out of range
+        //wwt.settings.set_showConstellationLabels(false) // can't make xml request
+        wwt.settings.set_constellationBoundryColor('#0000ff')
+        wwt.settings.set_constellationFigureColor('#ff0000')
         wwt.settings.set_constellationSelectionColor('#ffff00')
         wwt.settings.set_showCrosshairs(true)
+        wwt.settings.set_crosshairsColor('white')
```
astrofrog: It looks like one of the tests is failing because this should be ``'#ffffff``

astrofrog:
@ojustino - there are still some failures in `test_core.py` - can you make sure:

```
pytest pywwt/tests/test_core.py
```

passes locally for you?

ojustino:
The tests are passing locally for me now; I had to line some things up between `core.py` and `wwt.html`.

astrofrog:
Looks good, thanks!

PR #75, opened 2018-03-23
Make it easier to diagnose CI issues

astrofrog:
This includes a change that makes it easy to download the actual and expected images locally - basically there will now be a script printed in the log in case of failure that when copied/pasted locally will allow the actual and expected images to be re-generated (rather than trying to reproduce the exact same failure locally).

### DIFF
--- a/pywwt/tests/test_qt_widget.py
+++ b/pywwt/tests/test_qt_widget.py
@@ -73,12 +73,51 @@ class TestWWTWidget:
         wait_and_check_output(1, capsys)
 
 
+# The following is a template for a script that will allow developers who see
+# a failure in CI to re-create the files that were generated in the
+# continuous integration easily.
+REPRODUCABILITY_SCRIPT = """
+################################################################################
+# Export the images that were generated in the continuous integration for pywwt.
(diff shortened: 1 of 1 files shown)


astrofrog:
Example script from this run:

```
################################################################################
# Export the images that were generated in the continuous integration for pywwt.
# Just copy and paste all the code between here and '# End of script' into a
# local file and run it with Python. You can then check if the differences
# make sense, and if so, update the expected images.
import base64
expected = base64.b64decode('iVBORw0KGgoAAAANSUhEUgA...')
with open('expected.png', 'wb') as f:
    f.write(expected)
actual = base64.b64decode('iVBORw0KGgoAAAANSUhEUgAAAZAAAAGQ...')
with open('actual.png', 'wb') as f:
    f.write(actual)
# End of script
################################################################################
```

(I truncated the long strings)

And images generated:

**Expected**

expected

**Actual**

actual

This shows the planets are now real-size and the crosshair is gone.

astrofrog:
I'm going to merge this since it just adds a diagnostic tool

PR #81, opened 2018-04-06
Keyboard-controlled zoom

ojustino:
`z` to zoom in, `x` to zoom out. Any suggestions about zoom speed or smoothness?

There's still some debugging code in case I need to make changes, but this should work in both the Qt and Jupyter versions. Qt's web browser's standards aren't up-to-date as those of Chrome/Firefox/etc., so I had to add backwards compatibility to accommodate for that.

Now that I've figured out the problem with Qt, it seems like it wouldn't be difficult to add other keyboard controls.  Would that be useful?

### DIFF
--- a/docs/shortcuts.rst
+++ b/docs/shortcuts.rst
@@ -3,38 +3,55 @@
 Keyboard shortcuts
 ==================
 
-In both widgets, viewer navigation with the mouse follows typical conventions --
-click and drag up or down to move vertically; click and drag left or right to
-move horizontally; mouse wheel or pinch/two-finger scroll on a touchpad to zoom.
+In both the Qt and Jupyter widgets for WWT, viewer navigation with the mouse
+follows typical conventions -- click and drag up or down to move vertically;
+click and drag left or right to move horizontally; mouse wheel or
+pinch/two-finger scroll on a touchpad to zoom.
 
-In the Qt widget, it is also possible to navigate the viewer via keyboard
-shortcuts that are further explained below.  
+In the Qt widget (and the Jupyter widget in some browsers), it is also possible
+to navigate the viewer via keyboard shortcuts that are further explained below.
+ 
 
 Basic movements
 ---------------
 
 Imagining that the viewer's movements correspond with `an aircraft's principal
 axes of rotation <https://en.wikipedia.org/wiki/Aircraft_principal_axes>`_, you
-can perform two of the basic movements with the arrow keys alone:
+can perform two of the basic movements with the following keys:
 
-``up``, ``down``: pitch
-``left``, ``right``: yaw
+.. list-table::
+   :widths: auto
+
+   * - ``j`` or ``l``
+     - yaw
+   * - ``i`` or ``k``
+     - pitch
 
 To zoom, use:
 
-``z``: zoom in
-``x``: zoom out
+.. list-table::
+   :widths: auto
+
+   * - ``z``
+     - zoom in
+   * - ``x``
+     - zoom out
 
 Modified movements
 ------------------
 
-``alt`` allows you to access the third axis of rotation:
-``alt`` + ``left``, ``alt`` + ``right``: roll
+Use ``alt`` in combination with the directional keys to unlock more axes of
+movement:
 
-In individual planet modes, ``alt`` + ``up`` or ``down`` tilts the body
-vertically.
+.. list-table::
+   :widths: auto
+
+   * - ``alt`` + ``j`` or ``l``
+     - roll
+   * - ``alt`` + ``i`` or ``k``
+     - tilt vertically (individual planet modes only)
 
 Finally, adding ``shift`` to any of the above keys or combinations acts as a
 fine adjustment switch, performing the original action more slowly. For example,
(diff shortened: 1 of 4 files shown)


philrosenfield:
Awesome! 

I think these would be great in order of priority (but not suggesting you need to implement them all):
1. arrow keys mapping to camera or view
1. "fine adjust" which would be smaller increments of any key, and I'd suggest shift + that key.
1. "roll" mapping to camera or view 
1. planet "pitch" on the WWT Web client is ctrl +  mouse up/down (only in planet mode/solar system + planet focus)

For the last two, see https://en.wikipedia.org/wiki/Aircraft_principal_axes#/media/File:Flight_dynamics_with_text.png. These motions are all possible in the WWT web client.

Other possibly useful, but also possibly beyond the scope:
- some keystroke for a wwt.render call (ala print screen)
- map numbers to an ordered list of available_viewmodes or available_layers depending on the current view mode, e.g., 1=sun 2=mercury, etc. or 1=radio

ojustino:
New additions are movement with the arrow keys, the shift key as the fine adjust, and smoother movement/zooming overall.

ojustino:
`shift` is the fine adjust key. The full list should be:

`up`, `down`: pitch
`left`, `right`: yaw
`ctrl` + `left`, `ctrl` + `right`: roll
`ctrl` + `up`, `ctrl` + `down`: tilt (in planet mode)
`z`: zoom in
`x`: zoom out
`shift` + any of these keys/combinations: same function, but more slowly

philrosenfield:
Oh, shift made no difference for me, it works for you?

ojustino:
It works for me. It's most noticeable when zooming in or out, but maybe I should make the difference more pronounced?

ojustino:
A reminder to check this PR for issues and merge it if everything is up to par.

astrofrog:
@ojustino - can you rebase this PR?

Also when I try it locally I get:

```
Uncaught TypeError: Cannot set property 'shiftKey' of undefined (level=2, line_number=163, source_id=http://127.0.0.1:8000/data/wwt.html)
```

On MacOS X, control left/right isn't a good idea as it is already an existing shortcut to change workspace. What about using command- and shift- as the two modifiers?

ojustino:
Are you trying it in Qt or in a browser? I'm not able to reproduce that error using Qt in Linux. I can't test in-browser because of HTTPS issues with the SDK. 

Command is a hotkey on Windows, so that wouldn't work either. `alt` + left/right arrow would just function as a back/forward button in the browser when using a Jupyter notebook, right?

Maybe another solution is using i/j/k/l instead of the up/left/down/right arrows. I'm not too familiar with Macs; are those system-wide hotkeys when combined with `ctrl`? From what I've searched so far, they don't seem to be.

astrofrog:
I'm using Qt on Mac. If you don't have access to a Mac I can try and debug on my side.

ojustino:
I don't. Also let me know if `ctrl` + i/j/k/l work as the roll/tilt keys.

I'll also remember to write about these in the docs once we fix the error and decide on the proper keys.

astrofrog:
Actually the arrow keys are problematic in Jupyter as the up/down keys already mean something in a browser context, so maybe going with your idea of ijkl would make more sense. Could you easily change this?

astrofrog:
@ojustino - ok sounds good - it would be great if you could change to ijkl anyway since it works for me in the browser, but arrow keys already have meaning so there are multiple things happening at the same time.

docs/shortcuts.rst
```diff
@@ -0,0 +1,40 @@
+.. _layers:
+
+Keyboard shortcuts
+==================
+
+In both widgets, viewer navigation with the mouse follows typical conventions --
+click and drag up or down to move vertically; click and drag left or right to
+move horizontally; mouse wheel or pinch/two-finger scroll on a touchpad to zoom.
+
+In the Qt widget, it is also possible to navigate the viewer via keyboard
```
astrofrog: In the Qt widget, and in some browsers the Jupyter widget

docs/shortcuts.rst
```diff
@@ -0,0 +1,40 @@
...
+``alt`` allows you to access the third axis of rotation:
+``alt`` + ``left``, ``alt`` + ``right``: roll
+
+In individual planet modes, ``alt`` + ``up`` or ``down`` tilts the body
+vertically.
+
+Finally, adding ``shift`` to any of the above keys or combinations acts as a
+fine adjustment switch, performing the original action more slowly. For example,
+``shift`` + ``z`` zooms in at a slower pace, and ``shift`` + ``alt`` + ``left``
+gives you a slower roll motion.
```
astrofrog: I think it might be easier to simply list the shortcuts in a table - see:

http://docutils.sourceforge.net/docs/user/rst/quickref.html#tables

For how to do this in rst. Then maybe under the table you can keep the paragraph about the shift key to avoid repeating everything.

astrofrog:
Also one last request - please add an entry in the ``CHANGES.rst`` file

ojustino:
I wasn't sure what you meant by listing the shortcuts in tables, but I think it's useful to have the short commentary in between, especially with the explanations of the axes of movement.

PR #98, opened 2018-04-13
Moved solar system settings to proxy object

ojustino:
Is it OK that tab-completion also shows the inherited methods from traitlets' `HasTrait` along with the solar system traits?

Edits to the docs are upcoming.

### DIFF
--- a/docs/views.rst
+++ b/docs/views.rst
@@ -2,17 +2,18 @@ Switching views
 ===============
 
 New in release 0.4.0 is the ability to toggle between modes in the same manner 
-as in the WorldWide Telescope Web Client by using the ``set_view`` method. 
-Available modes include sky, planet, solar system, Milky Way, universe, and 
-panorama. The rest of the documentation is based on the default sky mode, so 
-here we discuss what makes the others different.
+as in the WorldWide Telescope Web Client by using the 
+:meth:`~pywwt.BaseWWTWidget.set_view` method. Available modes include sky, 
+planet, solar system, Milky Way, universe, and panorama. The rest of the 
+documentation is based on the default sky mode, so here we discuss what makes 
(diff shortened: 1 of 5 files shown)


ojustino:
I edited the docs to to reflect the proxy object.

I also enabled solar system lighting after testing it and being able to toggle it without issue -- both true and false worked. Are you two (cc @philrosenfield) still having problems with it?

docs/views.rst
```diff
@@ -73,11 +70,11 @@ The Universe view zooms all the way out to the extent of the observed universe:
...
 
-.. note:: If you're ever lost inside a mode, the ``reset_view`` method 
-          backtracks to the mode's initial state and allows you to start your
-          exploration over again.
+.. note:: If you're ever lost inside a mode, backtrack to its initial state by
+          using the :meth:`~pywwt.BaseWWTWidget.reset_view` method.
 
 Panorama view
 -------------
+**Keep or no?**
```
astrofrog: Is there any harm in keeping this for now?

pywwt/ss_proxy.py
```diff
@@ -0,0 +1,81 @@
+from astropy import units as u
+from traitlets import HasTraits, observe, validate, TraitError
+
+from .traits import Bool, Int
+
+class SolarSystem(HasTraits):
+
+    def __init__(self, BaseWWTWidget):
```
astrofrog: I would avoid using ``BaseWWTWidget`` here as that is the class name (not the instance) - instead use ``base_wwt_widget``

docs/views.rst
ojustino: I don't think so. My knowledge of it is just incomplete because I'm not sure what the mode-specific controls are.

Issue #68, opened 2018-01-29
Add text annotations


philrosenfield:
It would be nice to add text with the other annotations for example, following[ matplotlib.text](https://matplotlib.org/api/text_api.html)
```
wwt.add_circle(center, radius=radius)
wwt.add_text(string, xy=center, xytext=None, size=size, xycoords=canvas)
```
where xycoords would specify to tie the text to the WWT base layer or to the monitor.

astrofrog:
I agree, this would be handy!

philrosenfield:
FYI there is a workaround, but it's a little unclear to me how to trigger the text on and off with clicking/zooming (using qt):
`wwt.add_circle(center, radius=radius, label=somestring)`

ojustino:
The `label` and `hover_label` arguments for `wwt.add_circle()` are the standard way to add a label and make it visible. They also exist for `wwt.add_line()` and `wwt.add_polygon()`. I don't see how else to work with labels or otherwise add standalone text in the Web Control documentation, so is it currently possible access this ability?

I agree with Phil that it's hard to find what exactly triggers the label's appearance when using the viewer. It only shows up intermittently when I drag the view.

PR #102, opened 2018-04-24
Created option to display fields-of-view

ojustino:
I began with just Kepler; the function is called like `wwt.add_fov('k2')` with whichever other kwargs you'd like.

Should we save the telescope data in the repository or use HTTP requests to get it from where it's hosted? I included methods for both approaches.

Some questions about execution: should users be able to edit each individual polygon, or only all of them as a whole? Should there be a list of polygons like with `CircleCollection`? Should we make an "available telescopes" attribute so users can see what arguments `add_fov` takes?

**Note: The change in `wwt.html` should be reverted before this PR is merged. I made it so I could see annotations, and once we fix the issue with WebGL, I can switch the URL to the correct version.


astrofrog:
In terms of including the FOVs or accessing them remotely, I guess the question is how large the FOV files are going to be. In this case the JSON file is 40,000 lines long. Is all of that really needed or is there extraneous information we don't need?

ojustino:
Only taking what I need for `wwt.add_fov()`, I can cut the number of lines in K2's JSON file in half. The pretty-printed version of the file is 658 KB; as one line it goes down to 290 KB.

astrofrog:
Thinking about this some more, maybe what we could do to make it easier to add FOVs without having to release new versions of pyWWT each time would be to add the FOV files in a ``fov`` folder in the ``gh-pages`` branch, so that we don't have to rely on someone else hosting them - then we can make it so that when the user requests a FOV with name e.g. `k2``, it will automatically fetch the ``fov/k2.json`` file?

ojustino:
OK, I can open a new PR to add the FOV files in `gh-pages` and delete them from this one. Is there a way to programmatically access the path of a file on a different branch, or is using `requests` still the best we can do?

astrofrog:
@ojustino - you'll need to use requests to access the FOV files

ojustino:
As far as I can tell, changing the link in `wwt.html` to any of the previous versions no longer brings back the HTML5 version. (This includes http://worldwidetelescope.org/html5sdk/2.0/wwtsdk.js and https://raw.githubusercontent.com/WorldWideTelescope/wwt-web-client/master/webclient/sdk/wwtsdk.js among others.) 

Do you know why I'm getting the WebGL version no matter which link I use? The effect here is I'm not able to test if my call to requests to get the files gh-pages is working.

astrofrog:
@Justin - try changing the link to https://WorldWideTelescope.github.io/pywwt/wwtsdk.aspx - does this work?

astrofrog:
@ojustino - the default version we use now works properly. Can you change the URL back to what we had by default and see if things work now?

ojustino:
OK, I made the edits to `wwt.add_fov()` and added new instruments for HST and JWST. I can still work on Spitzer and SDSS, but I wanted to see what you thought about the current method of implementation.

ojustino:
As an aside, I think this is something that could use working text annotations (#68), otherwise most of these just look like assorted squares and rectangles. This is especially true when you a) generate FOVs from two or more instruments at the same time or b) use FOVs and regular annotations together.

ojustino:
Added IRAC from Spitzer.

Questions from before:
1) How does the implementation of this feature look so far?
2) Would it be a good idea to add text annotations to the FOVs so users can distinguish which polygon belongs to which instrument?

New questions:
3) Should users be able to rotate these fields of view?
4) How should we display SDSS, if at all? The point is to run each of the five rows of detectors over an object, but it would be hard to replicate that since a polygon is a static figure.

Also, I realize that the current code breaks near the poles since SkyCoord only accepts `dec` values in the range [-90, 90]. I'm working on a solution that will probably be in the next commit.

ojustino:
Added IRAC from Spitzer.

Questions from before:
1) How does the implementation of this feature look so far?
2) Would it be a good idea to add text annotations to the FOVs so users can distinguish which polygon belongs to which instrument?

New questions:
3) Should users be able to rotate these fields of view?
4) How should we display SDSS, if at all? The point is to run each of the five rows of detectors over an object, but it would be hard to replicate that since a polygon is a static figure.

Also, I'm aware that the current code breaks near the poles since `SkyCoord` only accepts `dec` values in the range [-90, 90]. I'm working on a solution that will probably be in the next commit.

pywwt/annotation.py
```diff
@@ -252,6 +254,155 @@ def _on_trait_change(self, changed):
...
+    """
+    A collection of polygon annotations. Takes the name of a pre-loaded 
+    telescope and displays its field of view.
+    """
+    # links for kepler. guide: https://keplerscience.arc.nasa.gov/k2-fields.html
+    # footprint: https://raw.githubusercontent.com/KeplerGO/K2FootprintFiles/master/json/k2-footprint.json
+
+    # should you be able to delete? change trait values?
+    # should there be a list of polygons like in CircleCollection?
+    # should self.available be an object so tab-completion is possible?
```
astrofrog: Yes to all! (with the tab completion being lower priority)

astrofrog:
> How does the implementation of this feature look so far?

This looks good so far. However, one thing we could consider is that rather than listing the available ones in this code, we simply read a JSON file from the gh-pages branch which can list the available footprints - this allows us to add footprints without releasing new versions of pywwt. The code should then not have anything specific to particular footprints.

> Would it be a good idea to add text annotations to the FOVs so users can distinguish which polygon belongs to which instrument?

Yes if possible - this could be an option for the user to choose? (whether to include a label or not)

> Should users be able to rotate these fields of view?

In general yes, except for field of views that are 'absolute' in that users can't put them anywhere they want - e.g. if we include the original Kepler field of view (not K2).

> How should we display SDSS, if at all? The point is to run each of the five rows of detectors over an object, but it would be hard to replicate that since a polygon is a static figure.

I'm not really sure if SDSS makes sense since most people just use the published mosaics - I think footprints will be more useful e.g. if planning observations (which one can't really do with SDSS). So having e.g. HST field of views is more useful than SDSS.

ojustino:
On the JSON file solution, I think it still might be necessary to update in order to add footprints because the code that makes polygons based on K2's footprint will probably not be applicable to others due to differences in structure. If that's the case, given the way `_gen_fov` currently works, we would then have to add new code specifically meant to parse the new footprint in `annotation.py`.

astrofrog:
@ojustino - it might make sense to pre-process the footprint files into a common format so that no changes are required in the pywwt code - that is, we can decide on a common format for all footprints, and just convert them all to that format.

ojustino:
That makes sense. We can consider how to do so once we've found/received more footprints.

pywwt/annotation.py
```diff
@@ -252,6 +255,218 @@ def _on_trait_change(self, changed):
...
+        # then, use trigonometry to rotate and undo the previous translation
+        rotated = concatenate((
+            SkyCoord(o_ra + r1*cos - d0*sin + swap0,
+                     o_dec + r1*sin + d0*cos, unit='deg'),
+            SkyCoord(o_ra + r0*cos - d0*sin + swap0,
+                     o_dec + r0*sin + d0*cos, unit='deg'),
+            SkyCoord(o_ra + r0*cos - d1*sin + swap1,
+                     o_dec + r0*sin + d1*cos, unit='deg'),
+            SkyCoord(o_ra + r1*cos - d1*sin + swap1,
+                     o_dec + r1*sin + d1*cos, unit='deg')))
```
astrofrog: @ojustino - here the offsets applied to RA should be ``/ cos(o_dec)``.

astrofrog:
I wonder if this could be written in a way that doesn't require HST or JWST-specific code. The main difference between those and e.g. the original Kepler FOV is that the Kepler FOV is absolute/fixed while other telescopes can point anywhere. So I think in the Python code we really should just distinguish between FOVs that can be moved to a specific RA/Dec and ones that don't, and do more work in terms of producing the JSON file so that it contains the full footprint.

We should then have a flag in the JSON that indicates if a footprint is relative or absolute. If absolute, then we can give the RA/Dec of the vertices (as done currently) and if relative, then we can give the offset in e.g. arcsec or degrees around the origin.

I think some of the math can be simplified in terms of translation and rotation. First, for the relative fields of view, we should rotate the offsets from the footprint *before* translating it to a given RA/Dec - this is because offsets in e.g. arcsec will be well behaved when rotating them, while things are a mess once we are in RA/Dec space. So step 1 is to simply do a 2D cartesian rotation of the offsets for relative footprints.

With the rotated offsets in place, you can then translate to the correct position by doing:

```
dec = dec0 + offsets
ra = ra0 + offsets / cos(dec)
```

Where ``ra0/dec0`` is the position were the user wants the footprint. Note that the last term is ``cos(dec)`` not ``cos(dec0)`` - that is, it's using the declination of each particular vertex in the footprint.

I think if you follow this model, the code will be a lot simpler and you should avoid the squashing issues you've been seeing.

pywwt/annotation.py
ojustino: Do you mean `o_ra` alone or the whole RA term?

astrofrog: I meant it should be ``o_ra + (r1*cos - d1*sin + swap1) / cos(o_dec)`` but in fact this code might be changed because of my comment below anyway.

ojustino:
In the latest commit, I made the edits suggested in the last comment. Some distortion issues still exist near the poles.

pywwt/instruments.py
```diff
@@ -0,0 +1,21 @@
...
+     # change link once new gh-pages PR is accepted!
+    
+    def __init__(self):
+        # instruments = requests.get('https://worldwidetelescope.github.io/pywwt/instruments.json').json()
+        self.dim = requests.get('https://raw.githubusercontent.com/WorldWideTelescope/pywwt/60facd7120af5a9c3f372c72617db1fa94dceab3/instruments.json').json()
+        self.available = list(self.dim.keys())
+
+    def __dir__(self):
+        # exclude extra dimensions of certain instruments (e.g. jwst_nircam)
+        shown = [inst for inst in self.available if inst != 'nircam_short']
```
astrofrog: Why the special case for nircam_short?

pywwt/annotation.py
```diff
@@ -252,6 +255,100 @@ def _on_trait_change(self, changed):
...
+
+        # test that pos matches what the user entered
+        if center and position == 'absolute':
+            raise ValueError('the given telescope does not take center coordinates')
+        elif not center and position == 'relative':
+            raise ValueError('the given telescope requires center coordinates')
+
+        # draw jwst nircam's inner panels first
+        if telescope[-6:] == 'nircam':
+            self._gen_fov('nircam_short', center, rot, **kwargs)
```
astrofrog: Why is this special case needed?

pywwt/annotation.py
```diff
@@ -252,6 +255,100 @@ def _on_trait_change(self, changed):
...
+
+        for panel in dimensions:
+            ras = (panel[0] * u.deg).value
+            decs = (panel[1] * u.deg).value
+            ra_r, dec_r = self._rotate(ras, decs, rot)
+
+            if telescope == 'k2':
+                ra_tr = ra_r; dec_tr = dec_r
```
astrofrog: If you want to write this on one line, I think ``ra_tr, dec_tr = ra_r, dec_r`` would be better.

pywwt/annotation.py
```diff
@@ -252,6 +255,100 @@ def _on_trait_change(self, changed):
...
+            else:
+                center_ra = center.ra.to(u.deg).value
+                center_dec = center.dec.to(u.deg).value
+
+                dec_tr = [center_dec + dec for dec in dec_r]
```
astrofrog: Does ``dec_tr = center_dec + dec_r`` not work?

pywwt/annotation.py
```diff
@@ -252,6 +255,100 @@ def _on_trait_change(self, changed):
...
+                # scale RA by dec (due to polar contraction of spherical coords)
+                ra_tr = [center_ra +
+                         ra / np.cos((dec_tr[i] * u.deg).to(u.rad)).value
+                         for i, ra in enumerate(ra_r)]
```
astrofrog: Likewise, can you not write ``ra_tr = center_ra + ra / np.cos((dec_tr[i] * u.deg).to(u.rad)).value``?

Note that the ``to(u.rad)`` is actually not needed, as ``np.cos`` will correctly recognize the degree unit

pywwt/annotation.py
```diff
@@ -252,6 +255,100 @@ def _on_trait_change(self, changed):
...
+            raise ValueError('the given telescope requires center coordinates')
+
+        # draw jwst nircam's inner panels first
+        if telescope[-6:] == 'nircam':
+            self._gen_fov('nircam_short', center, rot, **kwargs)
+
+        for panel in dimensions:
+            ras = (panel[0] * u.deg).value
+            decs = (panel[1] * u.deg).value
+            ra_r, dec_r = self._rotate(ras, decs, rot)
```
astrofrog: Rotations should only happen if ``position='relative'``?

pywwt/annotation.py
```diff
@@ -252,6 +255,100 @@ def _on_trait_change(self, changed):
...
+        # now, we rotate shapes at the origin before translating
+        rot = rot.to(u.rad).value
+        cos = np.cos(rot); sin = np.sin(rot)
+        ra_rot = []; dec_rot = []
+
+        i = 0
+        while i < len(ras):
+            ra_rot.append(ras[i]*cos - decs[i]*sin)
+            dec_rot.append(ras[i]*sin + decs[i]*cos)
+            i += 1
```
astrofrog: I don't think the loop should be needed? That is, I think the following should work:

```python
ra_rot = ras * cos - decs * sin
dec_rot = ras * sin + decs * cos
```

pywwt/core.py
```diff
@@ -435,6 +437,28 @@ def add_line(self, points=None, **kwargs):
             line.add_point(points)
         return line
 
+    def add_fov(self, telescope, center=None, rotate=0 * u.rad, **kwargs):
+        """
+        Add a telescope's field of view (FOV) to the current view.
```
astrofrog: It would be good to mention how to find out the list of available FOVs.

pywwt/annotation.py
ojustino: `dec_r` is a list of decs for each corner of a panel, so that's why I have to do a list comprehension here. (Also, if I make `dec_r` a length-4 `SkyCoord` and try to add `center_dec`, I get an error saying you can't add a `Quantity` or `int` to a `SkyCoord`.)

I need lists instead of SkyCoords because SkyCoord entries seem to be immutable (`'SkyCoord' object does not support item assignment`) and I need to be able to adjust decs and flip RAs for the `dec` check on line 314.

pywwt/annotation.py
ojustino: The same comment from above on needing lists and thus list comprehensions applies here.

For `np.cos`, I see now that degrees work. The documentation says radians only, but I guess that's not the case when the argument is an Astropy quantity.

pywwt/annotation.py
ojustino: I have the same comment on using lists here as on lines 306 and 308. 

I considered using `np.array` so I wouldn't need a loop, but the while loop ended up being 2-5 times faster when I measured speeds with `%timeit`. This makes a difference with big footprints like K2, which takes about a minute to draw right now.

pywwt/annotation.py
ojustino: `rot` is 0 by default, and the plan with the checks on lines 285 and 287 was to prevent the case where an FOV with `position='absolute'` carries different rotation value besides 0. 

I see that this isn't currently the case, but if it was, then `self._rotate(ras, decs, rot)` where `rot` is 0 would just return `ras` and `decs`. That being the case (and for simplicity's sake), I figured running everything through `self._rotate()` was no harm no foul.

Does that sound OK, or would you rather not call the function if no rotation is necessary? I think it comes down to whether adding another `if/else` condition is more time efficient than running `self._rotate()`.

pywwt/annotation.py
ojustino: I got confused earlier about where I was taking values. Long story short, I can do array-wise operations now so I don't need the list comprehensions anymore.

pywwt/annotation.py
astrofrog: I think we should just avoid calling the function at all if possible, for readability

pywwt/core.py
ojustino: Done.

astrofrog:
@ojustino - it looks like this will need to be rebased

pywwt/annotation.py
ojustino: I missed this reply. So you're saying don't call `self.rotate()` unless `position == relative` and `rot != 0`?

ojustino:
I think I rebased correctly -- in the header of `core.py`, we only want to import `LayerManager` (and not `TableLayer`), right?

If that and the other changes are correct and everything all right, please remember to merge #120 before this PR so I can change the link in `instruments.py` here.

ojustino:
I wrote preliminary documentation for this method.

Issue #103, opened 2018-06-20
Allow colors to be set to unicode in Python 2


astrofrog:
In Python 2, this fails with unicode strings:

```
        if isinstance(value, str) or (isinstance(value, tuple) and len(value) == 3):
```

(this is in the color trait code). We should use ``isinstance(value, six.string_types)`` instead of ``isinstance(value, str)`` so that unicode strings work.

ojustino:
Is this one fixed now?

astrofrog:
Yes it should be, thanks!

PR #105, opened 2018-06-26
Allowed unicode colors for Python 2

ojustino:
Fix for #103.

### DIFF
--- a/pywwt/traits.py
+++ b/pywwt/traits.py
@@ -79,7 +79,7 @@ class Color(TraitType):
             self.__doc__ = self.help
 
     def validate(self, obj, value):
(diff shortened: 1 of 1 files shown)


pywwt/traits.py
```diff
@@ -79,7 +79,7 @@ def __init__(self, *args, **kwargs):
             self.__doc__ = self.help
 
     def validate(self, obj, value):
-        if isinstance(value, str) or (isinstance(value, tuple) and len(value) == 3):
+        if isinstance(value, six.string_types) or (isinstance(value, tuple) and len(value) == 3):
```
astrofrog: I think the import for ``six`` is missing? Just add ``import six`` and include ``six`` in the ``install_requires`` list in ``setup.py``.

PR #112, opened 2018-08-06
Wrote branch-specific README

ojustino:
If there is more that should be added, please let me know.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,0 +1,7 @@
+About
+-----
+
+This branch of pyWWT hosts data files for sky surveys. At the moment, we only 
+offer the fields of the K2 campaign. If you have or are aware of files for 
+campaigns of other telescopes, please contribute them and we'll include them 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -0,0 +1,7 @@
+About
+-----
+
+This branch of pyWWT hosts data files for sky surveys. At the moment, we only
```
astrofrog: Maybe say 'footprints'? Note that this branch also stored the JS files required by PyWWT, so it might be worth mentioning.

ojustino: "Footprints" in place of "data files"?

astrofrog: Maybe it should say 'This branch of the PyWWT repository hosts data files needed by the pywwt package, including instrument footprints as well as javascript files'.

ojustino:
OK, I made a revision in the latest commit.

PR #120, opened 2018-08-23
Created JSON file with details of all instruments

ojustino:
This is file to read for the `add_fov` method, as mentioned in #102.


README.rst
```diff
@@ -2,7 +2,7 @@ About
...
 This branch of the pyWWT repository hosts data files needed by the pywwt 
-package, including instrument footprints and the JavaScript files that link the 
-package to the web client. The K2 campaign is the only footprint we host at the 
-moment, so if you can point us to files for other sky surveys, please let us 
-know and we'll include them here.
\ No newline at end of file
+package, including footprints and dimensions of instruments and the JavaScript 
+files that link the package to the web client. The K2 campaign is the only 
+footprint we host at the moment, so if you can point us to files for other sky 
+surveys, please let us know and we'll include them here.
```
astrofrog: @ojustino - does this description need updating now that we have more than just K2?

ojustino: K2 is the only one that displays a mosaic of annotations with absolute positions, which I thought was what we meant when we talk about "footprints." If that's the case, none of the other ones we have so far qualify.

PR #133, opened 2018-11-01
Made wwt_json_api.js function compatible with Qt

ojustino:
A one line change that prevents error read-outs for older versions of Qt.


ojustino:
That's a good call; I made that change.

Issue #125, opened 2018-10-18
Implement removal of layers


astrofrog:
We should implement a ``layer.remove()`` method that removes layers from the visualization.

philrosenfield:
+1 APLpy has this, and I’ve found it awkward (last time I’ve used it) to remove a layer with multiple plotted. Especially for notebook and lab users, if the layer isn’t saved as a variable, I couldn’t remove it. 

I’m sorry to not have had time to dig into the code more deeply before suggesting this, but do layers get added to a list that can always be accessed?

astrofrog:
@philrosenfield - not yet, see https://github.com/WorldWideTelescope/pywwt/issues/132 :)

astrofrog:
It looks like layer removal doesn't work at the moment with the JS API:

```
wwtlib.LayerManager.deleteLayerByID(layer.id)
```

doesn't seem to do anything - it removes the layer from the manager, but the layer is still visible. I also tried ``layer.cleanUp`` but no luck.

astrojonathan:
The functionality is already there, you just have to use all three parameters, if you skip the last two parameters it defaults as false and does nothing. Since the WWT layer universe is a multiverse (one for ambient/editing tours, one for playing tours) the delete can have different effects on the current state.

DeleteLayerByID(Guid ID, bool removeFromParent, bool updateTree)

Remove from parent will remove a tour from being referenced.. Must be true or the parental reference will still keep it alive.

Update tree will cause the UI to refresh to show the changes. 

In short when you use JavaScript to access c# projected methods, make sure you are not short-changing the parameter list. JavaScript is duck-typed and will let you call methods with too few too many or wrongly type parameters and hilarity ensues, so make sure you are checking the parameter list.

astrofrog:
Just confirming that fixes it - thanks!

PR #136, opened 2018-11-16
Add layer manager and way to remove layers

### LINKED ISSUE #132: Keep a reference to all existing layers
We should keep track of layers in the ``wwt`` object, for example as ``wwt.layers``. This will allow users to go and edit/remove layers even if they haven't kept a reference to them when creating them.

astrofrog:
Fixes #132

This does not solve #125 because it seems like the JS call doesn't have any effect so far, so we still need to fix that. But the Python API is in place.


astrofrog:
The AppVeyor build is broken in a way that is unrelated to this PR, so I'll go ahead and merge.

PR #145, opened 2018-12-05
Exposed a couple of features after SDK update

ojustino:
Namely, `set_solarSystemMinorOrbits` (orbit paths of non-featured moons, satellites, etc.) and `set_solarSystemStars` (background stars) in solar system mode. I also changed `solarSystemStars` to be on by default.

`set_solarSystemMinorPlanets` is also available now, but there are so many that it overloaded my RAM and crashed my session. In order to avoid this outcome for users, I've left it commented out.

### DIFF
--- a/pywwt/nbextension/static/wwt.html
+++ b/pywwt/nbextension/static/wwt.html
@@ -72,18 +72,18 @@
         wwt.settings.set_locationAltitude(0.0)
         wwt.settings.set_locationLat(47.633)
         wwt.settings.set_locationLng(122.133333)
-        //wwt.settings.set_solarSystemCosmos(false) // in source, don't think works
-        wwt.settings.set_solarSystemLighting(true) // in source, NOW works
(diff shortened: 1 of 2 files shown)


ojustino:
I should note that I tested all of the commented out settings in `wwt.html` using the SDK at http://www.worldwidetelescope.org/webclient/sdk/wwtsdk.min.js

(That version also has crosshairs again.) Once the latest changes to the SDK are staged and pushed, we should update our version of `wwtsdk.js` on the `gh-pages` branch.

astrofrog:
@ojustino - I've copied over the latest staged version to gh-pages. Can you update the URL for the JS and rebase this pull request? (to pick up the latest fixes to tests)

ojustino:
I only tested with the URL at worldwidetelescope.org, but the one in the commit still references the file in `gh-pages`. The features now work with the `gh-pages` SDK, too.

I rebased, so this PR should be good to go.

Issue #80, opened 2018-04-06
Find minimum version requirements for Jupyter/notebooks


ojustino:
`jupyter`, `jupyter_client`, `jupyter_console`, `jupyter_core`, `jupyterlab`, `jupyterlab_launcher`, and `notebook` should be somewhere within the following ranges:

    jupyter:             1.0.0-py36h9896ce5_0     --> 1.0.0-py36_4        
    jupyter_client:      5.1.0-py36h614e9ea_0     --> 5.2.3-py36_0        
    jupyter_core:        4.3.0-py36h357a921_0     --> 4.4.0-py36h7c827e3_0
    jupyterlab:          0.27.0-py36h86377d0_2    --> 0.31.12-py36_0      
    jupyterlab_launcher: 0.4.0-py36h4d8058d_0     --> 0.10.5-py36_0       
    notebook:            5.0.0-py36h0b20546_2     --> 5.4.1-py36_0

PR #149, opened 2018-12-11
Attempting version testing

ojustino:


.travis.yml
```diff
@@ -13,12 +13,14 @@ env:
...
     - PIP_DEPENDENCIES="sphinx-automodapi numpydoc sphinx_rtd_theme pytest-faulthandler codecov"
   matrix:
-    - PYTHON_VERSION=2.7
-    - PYTHON_VERSION=3.5
-    - PYTHON_VERSION=3.6
+		- PYTHON_VERSION=3.6 JUPYTERLAB_VERSION=0.27.0
+		- PYTHON_VERSION=3.6 JUPYTERLAB_VERSION=0.28.0
+		- PYTHON_VERSION=3.6 JUPYTERLAB_VERSION=0.29.0
+		- PYTHON_VERSION=3.6 JUPYTERLAB_VERSION=0.30.0
+		- PYTHON_VERSION=3.6 JUPYTERLAB_VERSION=0.31.0
```
astrofrog: Can you change these tabs back to spaces? I wonder if that could be causing issues with Travis?

ojustino: That and Travis fixing the issues they had on their side yesterday worked.

ojustino:
Maybe I can just link the results of each test after it's done so the build matrix isn't too large in any one commit. Here's the result for `jupyterlab`: https://travis-ci.org/WorldWideTelescope/pywwt/builds/467181132

ojustino:
`notebook` next: https://travis-ci.org/WorldWideTelescope/pywwt/builds/467224885?utm_source=github_status&utm_medium=notification

ojustino:
All of the Mac builds are failing. Do you know why that might be?

pkgw:
@ojustino Hmmm, I think the macOS failures are spurious, but it's a bit hard to tell. I tried rerunning one to see if anything changes.

I noticed that the setup stage has an error because the package `jupyterlab-launcher` is not available, but this is the case on the Linux builds too and it doesn't fail the overall CI there.

pkgw:
From my own testing, things seem to work if I change the `pip install` line in `.travis.yml` to be this:

```
  - pip install dist/*.tar.gz --verbose
```

Because Travis currently fails without an obvious error message, my guess is that this is a manifestation of [travis-ci#8920](https://github.com/travis-ci/travis-ci/issues/8920) somehow. Probably some earlier program is setting the shell's stdout to nonblocking mode. The bash shell has some code in place to avoid problems with this, but [as I have discovered, macOS bash doesn't](https://github.com/conda-forge/mpi4py-feedstock/pull/8#issuecomment-296841324).

My guess is that not using a compound shell command somehow avoids the nonblock problem?

pkgw:
Ah, OK, as a project maintainer I can push changes directly to the PR branch, so I've done so, Justin.

ojustino:
Thanks, @pkgw. Did the change lead to local success with macOS for you? It looks like everything is still failing on this branch.

pkgw:
Well, the CI got farther, at least ... dying on a later compound shell command. Let's try turning up the heat ...

pkgw:
@astrofrog @ojustino OK, I think we've made it to a non-magical error. When making the docs on macOS:

```
Exception occurred:
  File "/Users/travis/miniconda/envs/test/lib/python3.6/site-packages/qtpy/__init__.py", line 205, in <module>
    raise PythonQtError('No Qt bindings could be found')
qtpy.PythonQtError: No Qt bindings could be found
```

Hopefully this just indicates that some package needs to be installed?

astrofrog:
PyQt should be installed, it's just not importing for some reason. I'll investigate.

pkgw:
Revisited this and I believe with the latest force-push it should all work now.

pkgw:
Well, AppVeyor failed, but Travis worked!

For reference, here are example "actual" vs. "expected" images from where AppVeyor failed — it looks like the "actual" has lower resolution background imagery than the referene. Maybe that depends on whether the code can pull down the relevant tiles in time?

Anyway, that's all orthogonal to this PR, which I think should be mergeable now. @astrofrog @ojustino, what do you think?

actual
expected

ojustino:
This PR was made in response to #80, so we still need to go through and test the packages listed there to see what the minimum requirement for each is to run pyWWT successfully. I can get back to testing them on Travis like I was in earlier messages on this thread and see how that goes.

ojustino:
In the latest commit, I added some other tests we wanted to complete. We'll see which ones pass, and if there are no other errors, we should have our minimum version requirements soon.

ojustino:
OK, it appears that minimum versions are as follows (from [this Travis report](https://travis-ci.org/WorldWideTelescope/pywwt/builds/493885963?utm_source=github_status&utm_medium=notification)).

jupyter: 1.0.0
jupyter_client: 5.1.0
jupyter_console: 5.0.0
jupyter_core: 5.0.0
jupyterlab: 4.4.0
jupyterlab_launcher: 0.3.0
notebook: 5.0.0

@astrofrog does this solve #80?

astrofrog:
@ojustino - thanks! We should add a bit in

https://pywwt.readthedocs.io/en/stable/installation.html#id1

that says 'and if you want to use the Jupyter widget, ...'.

I don't think you need to specify all the packages you listed above, I think it should be sufficient to list notebook and jupyter and their minimum versions.

ojustino:
Now that #186 is merged, can we also merge this PR since its version of `.travis.yml` is more complete?

ojustino:
I have every test Travis test failing now. I tried backtracking to the commit with the last working version of `.travis.yml` and re-pushing it here, and all of the tests there failed as well. 

Do we have changes elsewhere that should affect Travis' success?

pkgw:
@ojustino Do you know how to click through and read the logs generated by the CI runs? It looks like the image comparison tests are failing — I'm going to guess that it's the same problem as I screenshotted before.

ojustino:
@pkgw I do, but I'm wondering why the failure has propagated over to Travis when it was contained to AppVeyor at the time you took the screenshots.

pkgw:
Oh, hm, the current CI failure is not quite the same. The "actual" images are all black, on both Windows and Linux. (The test suite isn't run on macOS.) I can't reproduce locally on my Linux machine. The same issue is showing up in #192.

PR #165, opened 2019-01-03
Further improvements to layers

astrofrog:
This implements the ability to specify whether marker size is relative to the screen or the world coordinates, as well the the marker type (although see https://github.com/WorldWideTelescope/wwt-web-client/issues/191 for a description of issues related to that).

This also adds a workaround for https://github.com/WorldWideTelescope/wwt-web-client/issues/189 which means that plotting relative to e.g. the Sun and bodies other than the Earth will work correctly when the altitude/distance is provided.

Finally, this exposes the option to show/hide points on the far side of an object (see https://github.com/WorldWideTelescope/wwt-web-client/issues/190 for what I think is an issue with the default setting of ``False``).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,14 @@
 
 - Fixed issues with zooming using trackpad/scroll wheel. [#166]
 
+- Added support for customizing the layer marker type (``marker_type``), the
+  option to specify whether the marker size is absolute or relative to the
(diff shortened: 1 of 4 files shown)


astrofrog:
Thanks! The Travis failure is unrelated, so I'll go ahead and merge this.

PR #166, opened 2019-01-03
Don't override default wheel behavior

astrofrog:
Since https://github.com/WorldWideTelescope/pywwt/pull/81, the two finger scroll results in a different behavior from the WWT web client (the axis direction is inverted) and the scrolling can be jumpy at times. Removing this block of code restores the original behavior. It seems this isn't needed for the keyboard shortcuts?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 0.5.0 (unreleased)
 ------------------
 
-- No changes yet.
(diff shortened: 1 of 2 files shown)


astrofrog:
I've now taken a different approach - I've renamed the move and wheel events to not interfere with the default events when using the mouse/trackpad.

PR #168, opened 2019-01-04
Remove post/pre scripts

astrofrog:
I think these are no longer needed, though we'll see based on the ``run_test.py``


astrofrog:
@pkgw - ok I've removed it, thanks!

pkgw:
So, are the commands needed to enable Jupyter(Lab) extensions not going to be needed going forward?  That would be nice if true ...

astrofrog:
This looks like it's good to go, so merging

PR #170, opened 2019-01-04
Added note about using Jupyter Lab

### LINKED ISSUE #169: Document how to set up widget in Jupyter Lab
We should make sure we properly document how to set up the widget for Jupyter Lab.

astrofrog:
@pkgw - this works for me, is this also what you had to do to get Jupyter Lab to work?

Fixes #169

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -22,6 +22,12 @@ release of pywwt using::
 This will automatically install pywwt, its `dependencies <Dependencies>`_, and
 will enable the Jupyter extension.
 
+If you want to use WWT inside Jupyter Lab, you may need to also run::
+
+    jupyter labextension install @jupyter-widgets/jupyterlab-manager
+
+if you have never used widgets in Jupyter Lab before.
+
 Installing pywwt with pip
 -------------------------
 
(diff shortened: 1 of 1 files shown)


pkgw:
I see that you already merged this, but it might help to also mention the `jupyter lab build` command. As far as I can tell, Jupyter will run it itself if needed, but errors might be caught sooner if users run it themselves. (Especially since the build command seems to farm out to a lot of subcommands: `npm`, `yarn`, `webpack` ... I'm not sure which of those are expected to be available on the system, and which are installed by other tools.)

pkgw:
Although ... I managed to wedge my install somehow when playing with this stuff.

If you run `jupyter labextension install pywwt`, it will apparently go to the NPM registry for pywwt ... which has an NPM entry, but it's a year old. The first command that gets run is `npm pack pywwt` and that packages up version 0.3.0: <https://www.npmjs.com/package/pywwt>.

Probably this is just something that happens if you fool around too much, but perhaps we should be taking care to update the NPM versions when making releases?

astrofrog:
@pkgw - thanks for catching that, I've published 0.5.1 on npm: https://www.npmjs.com/package/pywwt/v/0.5.1 - I'll update the docs along the lines you suggest too.

Issue #138, opened 2018-11-16
Color-coding of points in layers


astrofrog:
Similar to #137, it would be nice to be able to be able to define the color-coding of points in layers by specifying a column (currently ``colorMapColumn``) but also ``colorMapMin`` and ``colorMapMax`` such that:

```
colorFloat = (parseFloat(row[this.colorMapColumn]) - this.colorMapMin) / (this.colorMapMax - this.colorMapMin);
```

and then ``colorFloat`` would get mapped to a colormap. Until this is present in the JS, we will need to do this in Python and then update the data every time ``colorMapMin`` and ``colorMapMax`` are changed.

astrofrog:
Just for the record, this is how WWT currently handles layer colors:

```
            switch (this.get__colorMap()) {
              case 0:
                pointColor = colorLocal;
                break;
              case 3:
                if (this.get_colorMapColumn() > -1) {
                  pointColor = this._parseColor$1(row[this.get_colorMapColumn()], colorLocal);
                }
                else {
                  pointColor = colorLocal;
                }
                break;
              default:
                break;
            }
```

So for now we need to include a column with the list of colors (one for each point).

PR #183, opened 2019-02-15
Implement support for color and size encoding based on attributes

### LINKED ISSUE #137: Scaling of points in layers
The SDK has a ``sizeColumn`` that we expose in the data layer - this makes it possible to encode the size of points by one of the columns. There are different ``pointScaleType`` modes - however, none of them allow us to do something like:

```js
pointSize = (parseFloat(row[this.sizeColumn]) - this.sizeMin) / (this.sizeMax - this.sizeMin);
```

Having this mode is important in cases where we want to scale by a column where the actual values of the column need to be mapped to - but aren't - point sizes. An example would be magnitudes for which there will be negative values. Without this mode, we have to pre-compute a size column, and then re-pass all the data to JS if we decide we want to change the min/max size. With the above mode, we could easily change the scaling efficiently without having to add dummy columns.

astrofrog:
This implements proper support for specifying ``size_att`` and ``cmap_att`` by computing the sizes and colors Python-side using min/max values (specified by ``size_vmin``, ``size_vmax``, ``cmap_vmin``, and ``cmap_vmax``), since such a behavior is not supported for now WWT-side. This has a performance hit because whenever we update one of the limits or the attributes we need to send the table from Python to JS. The ideal solution is to implement support for this kind of scaling WWT-side, which I'll try and work on in future (but the idea here is to get the Python API functional even if performance isn't optimal).

Things that remain to be done here:

* [x] Add image tests
* [x] Figure out how to work around https://github.com/WorldWideTelescope/wwt-web-client/issues/203
* [ ] <s>Vectorize to_hex calculation</s> - let's address this if it becomes a performance concern
* [x] Add documentation
* [x] Add changelog entry

Fixes #137 and #138


astrofrog:
Note: I've included the fix from https://github.com/WorldWideTelescope/wwt-web-client/pull/207 in the SDK we serve from GitHub to avoid holding this up.

astrofrog:
For some reason the image in the CI comes out blank...

pywwt/layers.py
```diff
@@ -162,11 +170,23 @@ class TableLayer(HasTraits):
...
     size_scale = Float(10, help='The factor by which to scale the size of the points').tag(wwt='scaleFactor')
 
+    # NOTE: we deliberately don't link size_att to sizeColumn because we need to
+    # compute the sizes ourselves based on the min/max and then use the
+    # resulting column.
+    size_att = Unicode(help='The column to use for the size')
+    size_vmin = Float(None, allow_none=True)
+    size_vmax = Float(None, allow_none=True)
+
+    # NOTE: we deliberately don't link size_att to sizeColumn because we need to
```
pkgw: Should this comment reference `cmap_att` and something like `cmapColumn`?

pywwt/nbextension/static/wwt_json_api.js
```diff
@@ -234,14 +234,9 @@ function wwt_apply_json_message(wwt, msg) {
...
-
-      // FIXME: workaround for the fact that at the moment, WWT appears
-      // to only refresh if the color is changed. So we change to black then
-      // back (https://github.com/WorldWideTelescope/wwt-web-client/issues/192)
-      color = layer.get_color();
-      layer.set_color(wwtlib.Color.fromHex('#000000'));
-      layer.set_color(color);
+      // Use updateData instead of loadFromString here since updateData also
+      // takes care of cache invalidation.
+      layer.upadteData(csv, true, true, true)
```
pkgw: Typo here.

astrofrog: @pkgw - actually not a typo here, it's a WWT typo ;) https://github.com/WorldWideTelescope/wwt-web-client/pull/205

pywwt/layers.py
astrofrog: Fixed

Issue #123, opened 2018-10-18
Automatically zoom on planet when using set_track_object


astrofrog:
At the moment, when using ``set_track_object`` in solar system mode, one still has to zoom in manually to the planet. I believe there is a function ``gotoTarget`` that we may be able to use to do this, but this requires a ``Place`` object and I haven't had a chance to find out how to get such an object for a given planet.

astrofrog:
Thanks @astrojonathan! I'll re-open this until we implement support for it in PyWWT

ojustino:
@astrojonathan Thank you for the assist; I nearly have this fixed. One more question:

The `Place.create` + `gotoTarget` method works for all expected bodies except Earth. When I feed the string `Earth` to `gotoTarget`, I get this error: `TypeError: undefined is not an object (evaluating 'raDec.RA') (line_number=35853, source_id=https://worldwidetelescope.github.io/pywwt/wwtsdk.js)`.

This is in the `get_params` method of `Place`, which calls `Planets.getPlanetLocation`, which in turn calls `Planets.getPlanetIDFromName`. 

`Earth` is one of the entries in the `switch`/`case` statement of `getPlanetIDFromName`, so this method returns `19`. 

That value is passed back to `getPlanetLocation` to use as an argument when calling `AstroCalc.getPlanet`. In the latter method, I see that there's no `case` for when the planet ID is 19 (corresponding with Earth).

Is that intentional, and if so, how can I work around it?

ojustino:
@astrojonathan This comment is lower priority than the previous one, but `gotoTarget(place, noZoom, instant, trackObject)` is also not working as expected -- I would like a smooth scroll to the planet, but the view changes instantly whether `instant` is true or false.

I tried using the `.aspx` version of the SDK and got the desired scrolling zoom, so is this something that doesn't work normally with WebGL?

Issue #177, opened 2019-01-16
Add the ability to center on specific lon/lat for planetary data


astrofrog:
The ``center_on_coordinates`` method only really works for the Sky view, but it'd be nice to have something like this for lon/lat on a planet for instance. I'm pretty sure this exists in the WWT API, it's just a case of exposing it here.

PR #186, opened 2019-02-24
Added auto-zoom to solar_system.track_object

ojustino:
Addresses #123.  The changes in the first commit fix the issues I was having with both animated transitions and selecting Earth.

I use `gotoRaDecZoom` for the Sun since it the zoom levels for `gotoTarget` appear to be immutable, and I thought it made the Sun a little small compared to how the other planets and moons appear. `gotoTarget` appears to default to the behavior of the `gotoRaDecZoom` when the target is the Sun, anyway, so using `gotoRaDecZoom` gains you direct control over the zoom level.

If this is all good and there are no more changes to make, I can delete the comments before we merge this PR.

### DIFF
--- a/pywwt/nbextension/static/wwt_json_api.js
+++ b/pywwt/nbextension/static/wwt_json_api.js
@@ -184,33 +184,36 @@ function wwt_apply_json_message(wwt, msg) {
       wwt.setBackgroundImageByName(msg['mode']);
       wwt.setForegroundImageByName(msg['mode']);
       break;
-/*
-    case 'track_object':
-      wwtlib.WWTControl.singleton.renderContext.set_solarSystemTrack(msg['code']);
-      break;
-*/
+
     case 'track_and_zoom':
+
       var object = msg['obj'];
+      var instant = msg['inst'];
+
+      // Collect the arguments necessary to create a Place object
       var ra = wwt.getRA();
       var dec = wwt.getDec();
       var classification = wwtlib.Classification.solarSystem;
-      var constellation = wwtlib.Place._constellation;
+      var constellation;
       var imageType = wwtlib.ImageSetType.sky;
       //var imageType = wwtlib.ImageSetType.solarSystem;
       var scale = wwt.settings.get_solarSystemScale();
       var zoom; //= (.02 * scale + .08) / msg['zoom'];
-      // if imageType == wwtlib.ImageSetType.solarSystem, we need the equation above
-      // else, zoom is pre-set in gotoTarget if imageType == wwtlib.ImageSetType.sky
+      // if imageType == wwtlib.ImageSetType.solarSystem, use the equation
+      // elif imageType == wwtlib.ImageSetType.sky, gotoTarget pre-sets zoom
 
-      var place = wwtlib.Place.create(object, ra, dec, classification, 
+      // Create the Place object and focus it on the object in question
+      var place = wwtlib.Place.create(object, ra * 15, dec, classification,
                                       constellation, imageType, zoom);
       place.set_target(wwtlib.Planets.getPlanetIDFromName(object));
 
+      // Go to the place that was created (if not the Sun)
+      // Else, use the standard, Sun-centered method of centering on coords
       if (object != 'Sun') {
-        wwtlib.WWTControl.singleton.gotoTarget(place, false, msg['inst'], true);
+        wwtlib.WWTControl.singleton.gotoTarget(place, false, instant, true);
       }
       else {
-        wwt.gotoRaDecZoom(ra, dec, .00221*scale + .00552, msg['inst']);
+        wwt.gotoRaDecZoom(ra * 15, dec, .00221*scale + .00552, msg['inst']);
       }
 
       break;
--- a/pywwt/nbextension/static/wwt_json_api.js
+++ b/pywwt/nbextension/static/wwt_json_api.js
@@ -190,24 +190,28 @@ function wwt_apply_json_message(wwt, msg) {
       var object = msg['obj'];
       var instant = msg['inst'];
 
-      // Collect the arguments necessary to create a Place object
+      // First, collect the arguments necessary to create a Place object
+
+      // (record current ra/dec so they're preserved if the object is the Sun)
       var ra = wwt.getRA();
       var dec = wwt.getDec();
+      // (create other args for Place.create(), keeping unecessary ones blank)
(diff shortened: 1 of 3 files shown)


pywwt/nbextension/static/wwt_json_api.js
```diff
@@ -182,10 +182,36 @@ function wwt_apply_json_message(wwt, msg) {
       wwt.setBackgroundImageByName(msg['mode']);
       wwt.setForegroundImageByName(msg['mode']);
       break;
-
+/*
     case 'track_object':
       wwtlib.WWTControl.singleton.renderContext.set_solarSystemTrack(msg['code']);
       break;
+*/
+    case 'track_and_zoom':
```
astrofrog: Can you remove the old code?

pywwt/nbextension/static/wwt_json_api.js
```diff
@@ -182,10 +182,36 @@ function wwt_apply_json_message(wwt, msg) {
...
+      var object = msg['obj'];
+      var ra = wwt.getRA();
+      var dec = wwt.getDec();
```
astrofrog: I don't understand why the current RA/Dec are relevant. What do those mean in the context of the planet/body views? Is it related to the direction from which one is viewing the solar system? It would be good to add some comments here to explain it.

pywwt/nbextension/static/wwt_json_api.js
```diff
@@ -182,10 +182,36 @@ function wwt_apply_json_message(wwt, msg) {
...
+      var classification = wwtlib.Classification.solarSystem;
+      var constellation = wwtlib.Place._constellation;
```
astrofrog: This appears to use a private property, which is probably ok for now but it would make sense to ask whether this could be made public to avoid breakage in future?

pywwt/solar_system.py
```diff
@@ -75,15 +75,39 @@ def track_object(self, obj):
...
+                    'Callisto': 13, 'Ioshadow': 14, 'Europashadow': 15,
+                    'Ganymedeshadow': 16, 'Callistoshadow': 17,
+                    'Suneclipsed': 18, 'Earth': 19}'''
+
+        '''zooms = {'Sun': sun, 'Mercury': rock, 'Venus': rock, 'Mars': rock,
+                 'Jupiter': gas, 'Saturn': gas, 'Uranus': gas, 'Neptune': gas,
+                 'Pluto': rock, 'Moon': rock, 'Io': rock, 'Europa': rock,
+                 'Ganymede': rock, 'Callisto': rock, 'Ioshadow': rock,
+                 'Europashadow': rock, 'Ganymedeshadow': rock,
+                 'Callistoshadow': rock, 'Suneclipsed': sun, 'Earth': rock}'''
```
astrofrog: Is the above needed?

pywwt/nbextension/static/wwt_json_api.js
ojustino: I get them a) because RA and dec are arguments of `Place.create()`, and b) so RA/dec remain the same if you're tracking the Sun. (If you're not tracking the Sun, then the current RA/dec do not actually influence the final orientation of the object due to #177 -- as far as I can tell.)

Given that the current RA/Dec doesn't seem to matter when creating a `Place`, would you rather assign a fixed RA/dec for the Sun and just pass fixed, pre-selected RA/dec values to `Place.create()`?

pywwt/nbextension/static/wwt_json_api.js
ojustino: Removing for the next commit.

pywwt/nbextension/static/wwt_json_api.js
ojustino: I wanted to include it since `constellation` is an argument for `Place.create()`, but I logged it in the console and it turns out that its value is always `undefined`. 

I think I'll do what I did with `zoom` and create the variable (so we know what's what the argument means) but not assign it a value. That way there's no need to change the visibility of any properties in the SDK.

pywwt/solar_system.py
ojustino: No, I just kept it around in case it might be useful after the review. Removed for the next commit.

ojustino:
@astrofrog Thanks for taking a look. 

The non-instantaneity of the `instant` flag is not by design, but I didn't find how to get truly instantaneous behavior while using this method of focusing the viewer on a planet.

I removed the extra code. Please let me know what you think of my responses on what to do with `ra`, `dec`, and `constellation`.

pywwt/nbextension/static/wwt_json_api.js
astrofrog: I don't really have a strong preference, but I think that whatever you do you should add a comment to explain why you are doing it

pywwt/nbextension/static/wwt_json_api.js
astrofrog: Sounds good!

pywwt/nbextension/static/wwt_json_api.js
```diff
@@ -179,8 +179,37 @@ function wwt_apply_json_message(wwt, msg) {
...
+      var ra = wwt.getRA();
+      var dec = wwt.getDec();
+      var classification = wwtlib.Classification.solarSystem;
+      var constellation;
+      var imageType = wwtlib.ImageSetType.sky;
+      //var imageType = wwtlib.ImageSetType.solarSystem;
+      var scale = wwt.settings.get_solarSystemScale();
+      var zoom; //= (.02 * scale + .08) / msg['zoom'];
+      // if imageType == wwtlib.ImageSetType.solarSystem, use the equation
+      // elif imageType == wwtlib.ImageSetType.sky, gotoTarget pre-sets zoom
```
astrofrog: There is still some unused code here - can you remove anything that's not used?

ojustino: I put the extra code in a comment that describes a method we might use to give more flexibility to this function once we fix RA/Dec for planets. It would be nice to keep it handy so I don't forget what I found when searching through the SDK.

astrofrog:
@ojustino - could you rebase this? Is it ready for a final review?

ojustino:
@astrofrog Rebased and ready for review.

astrofrog:
@jsub1 - could you advise @ojustino on how to fix the serialization test that is failing? (related to the fact we no longer hard code the object IDs in Python)

Issue #175, opened 2019-01-11
Expose ability to have time-dependent layers


astrofrog:
Layers have the ability to define time-dependent markers, so we should expose that functionality in the layer API.

ojustino:
@astrojonathan showed me that the code to this lives here: https://github.com/WorldWideTelescope/wwt-web-client/blob/master/HTML5SDK/wwtlib/Layers/SpreadSheetLayer.cs

Specifically, we can use `BeginRange`, `EndRange`, `Decay`, and the boolean `TimeSeries` that turns on all of this functionality.

ojustino:
I'm finding that these settings are trapped by scope inside objects in the `wwtlib.SpreadSheetLayer` and `wwtlib.TimeSeriesLayer` classes and am not finding the way to edit them once a layer has been created. Is it only possible to set these attributes when initializing a layer? @astrojonathan

astrojonathan:
No these are settable at run-time, and in fact the web client binds to the active layer and sets these thru UI, so You can do the same.
What have you tried? They are accessible like any other property.

ojustino:
So when working through the SDK, I should be able to do something like `wwtlib.SpreadSheetLayer.set_beginRange(date)`? With that, I get `TypeError: undefined is not an object (evaluating 'wwtlib.SpreadSheetLayer.get_beginRange')`.

astrojonathan:
No. You set_timeSeries = true;  The "begin" and "end" ranges are internally computed from your data.
You litteally just need to set the columns to define what column number defines your begin and optionally end data, your decay an set the time series to true at it just works.
From that point, you can then loop thru the time range using the space time controller and watch the animation happen.

ojustino:
I'm not sure why it's not working so simply for me. The only editable attributes I get from `wwtlib.SpreadSheetLayer` (using `Object.keys()`) are `_getDatafromFeed$1`, `_executeQuery$1`, `parseDate`, `execlToDateTime`, `get__circleTexture$1`.

From what I see in the `SpreadSheetLayer` section of the SDK, `set_timeSeries` is inside an object called `SpreadSheetLayer$` that I can't access from outside; the errors always say it's undefined.

pkgw:
[...]

PR #187, opened 2019-03-04
Troubleshooting time series layer behavior

ojustino:
@astrojonathan Here is an environment where you can test pywwt directly to figure out what's going on. I've added `console.log` statements to the `wwt_json_api.js` file from earlier that should print messages to the console to let you know what's being passed to the SDK. The instructions to set it the time series test are as follows.

**Install the developer version of pywwt in terminal** (if you don't have it already):
```
git clone https://github.com/WorldWideTelescope/pywwt.git
cd pywwt
pip install -e .
```
**Clone this pull request branch to your machine**:
```
git checkout -b ojustino-layer-time master
git pull https://github.com/ojustino/pywwt-web.git layer-time
```
It's possible to run pywwt via script, but it might be better to troubleshoot interactively on this one, so **go to the `pywwt/nbextension/static` directory, open IPython on the command line, and then copy/paste as instructed from `time_test.py`**. 

A Qt window running the same `wwt.html` file we worked from last week will pop up once you've copied, pasted, and run the code block with the imports from `time_test.py`.


ojustino:
The first three of the new commits are focused on making what I hope are _small improvements in trait functionality_ by expanding the docstrings for TableLayer traits, making the error messages in `AstropyQuantity` and `Color` more informative, and resolving some JS TypeErrors that arose from the `wwt.*ground` attributes having a flag they shouldn't have.

@pkgw, **the most notable development is that I got time series layers to show up in the Qt version**. I was also able to get **decay** working, so, if the user wants, they can make it so points disappear after a specific amount of time. _(I also added `time_test.ipynb` to verify that things were still working for a Jupyter notebook in Firefox.)_

I hear the concern in #175 about **parsing dates from strings**, but I don't think there's another way to do it that works for the Qt version. My solution verifies that those strings all conform to the ISOT standard before passing them to WWT. ISOT is browser-independent and should hopefully then be friendly with WebKit, Chromium, and Firefox.

As for **time zones** in Qt, I used `datetime` to find the user's current time and add the offset between that and the current UTC time to values in the time column before passing the column to WWT. **The current drawback of this approach is Daylight Saving Time.** If the user's current DST status is not the same as it is for the dates in their time column, there's a one hour offset. (The test script for the Qt version illustrates this discrepancy since the events are set in January.) It's also assumed that users are providing UTC times, but I hope that can be expanded in the future.

**Other questions I have are marked in comments beginning with '`ISSUE`' in `layers.py`,** but that's the general summary. It would be great if @astrofrog could take a look and see how my implementation fits with how the rest of that file was set up.

pywwt/layers.py
```diff
@@ -236,40 +256,72 @@ class TableLayer(HasTraits):
...
+    # the resulting column.
+    cmap_att = Unicode(help='The column to use for the colormap '
+                       '(`str`)').tag(wwt=None)
+    cmap_vmin = Float(None, help='The minimum level of the colormap. Found '
+                      'automagically once cmap_att is set (`float`)',
+                      allow_none=True).tag(wwt=None)
+    cmap_vmax = Float(None, help='The maximum level of the colormap. Found '
+                      'automagically once cmap_att is set (`float`)',
+                      allow_none=True).tag(wwt=None)
+    cmap = Any(cm.magma, help='The Matplotlib colormap '
```
astrofrog: Why change the default? I thought viridis would make sense for consistency with matplotlib

pywwt/layers.py
```diff
@@ -236,40 +256,72 @@ class TableLayer(HasTraits):
...
+    far_side_visible = Bool(False, help='Whether markers on the far side '
+                            'of a 3D object are visible '
+                            '(`bool`)').tag(wwt='showFarSide')
+
+    # NOTE: we deliberately don't link time_att to startDateColumn here
+    # because we need to compute a new times column based on time_att before
+    # passing the result on to WWT
+    time_att = Unicode(help='The column to use for time (`str`)').tag(wwt=None)
+    time_series = Bool(False, help='Whether the layer contains time series '
+                       'elements (`bool`)').tag(wwt='timeSeries')
```
astrofrog: Do we need this separate option, or would it be sufficient to check that ``time_att`` is not None? (there's a way to get a Unicode trait to also accept None, check the traitlets docs as I have forgotten the exact option name)

pywwt/layers.py
```diff
@@ -236,40 +256,72 @@ class TableLayer(HasTraits):
...
+    decay = AstropyQuantity(16 * u.day, help='How long a time series point '
```
astrofrog: decay -> time_decay

pywwt/layers.py
```diff
@@ -285,6 +337,17 @@ def __init__(self, parent=None, table=None, frame=None, **kwargs):
         self.table = table
         self.frame = frame
 
+        # ISSUE: For some reason, WWT seems to prefer the that time column
+        # be 0th-10th/12th column in the table.
+        ### (varies with placement of latitude and longitude columns,
+        ###  which also need to be 0th-12th)
+        # So the current proposition is to limit table's max number of columns
+        ### (the max will change depending on how time_att is implemented)
+        max_cols = 13
```
astrofrog: I think rather than do this it would be better to check when ``time_att`` is set whether it is in within the range of columns is 0-12, and give an error if not explaining that this won't work for now. I've definitely tried to plot tables with more than 13 columns before and I don't think we want to constrain things in all the cases where time is not used.

Of course, it'd be even better if we can figure out what is causing this limitation :) Maybe @jsub1 has some ideas?

pywwt/layers.py
```diff
@@ -374,6 +442,37 @@ def _check_cmap(self, proposal):
...
+    def _check_time_att(self, proposal):
+        # Parse the time_att column and make sure it's in the proper format
+        # (string in isot format, astropy Time, or datetime)
+        col = self.table[proposal['value']]
+
+        if (isinstance(col, STR_TYPE)
+            or np.issubdtype(col.dtype, NP_STR_TYPE)):
+
+            col_list = col.tolist()
+            is_iso = all(iso_test.match(t) for t in col_list)
```
astrofrog: Note that you could also do the following if you want to avoid implementing the validation:

```python
try:
    Time(col, format='isot')
    return proposal['value']
except <exceptiontype>:
    raise ValueError('...')
```

And in future, we could then even start to support more string formats that astropy supports.

pywwt/layers.py
```diff
@@ -502,6 +601,62 @@ def _on_cmap_vmin_vmax_change(self, *value):
...
 
+
+    @observe('time_att')
+    def _on_time_att_change(self, *value):
+        # Convert time column to UTC so WWT displays points at expected times
+        ### ISSUE: Currently not robust to changes in DST.
+        # Needs GMT instead of UTC?
+
+        # The conversion only seems needed in Qt, so check the widget version
+        if type(self.parent).__name__.find('Qt') < 0:
```
astrofrog: What versions is this actually checking?

pywwt/nbextension/static/time_test.ipynb
```diff
@@ -0,0 +1,191 @@
+{
```
astrofrog: Is this file temporary?

pywwt/nbextension/static/time_test.py
```diff
@@ -0,0 +1,45 @@
+# copy/paste each in segments up to each comment
```
astrofrog: Same question, is this temporary?

pywwt/layers.py
ojustino: This was a test I was running that I didn't return to its original value. I'll restore viridis in the next commit.

pywwt/layers.py
ojustino: @pkgw and I were also talking about this. I don't have a strong opinion on it. My main goal was to try to replicate the way layer creation works in the code from the SDK, but we naturally have the latitude to change this in `pywwt` if we desire.

pywwt/layers.py
ojustino: You'll see this in the next commit.

pywwt/layers.py
ojustino: I'm trying to check whether we're in an instance of `WWTJupyterWidget` or `WWTQtClient` since Chrome/Firefox and WebKit handle `Date` objects differently. Does this not work for you?

Ideally, this step won't be necessary after we iron out the current issues with time zones.

pywwt/nbextension/static/time_test.ipynb
ojustino: Yes. I made it so we could all work from the same test and see if my commits work for everyone or are specific to my set-up (especially with web browsers). I'll remove it once we have a working solution.

pywwt/nbextension/static/time_test.py
ojustino: Same response as above.

pywwt/layers.py
ojustino: @jsub1 You're right, it's a parsing issue with the `place` column, not the number of columns in the table. I'll take this part out in the next commit.

pywwt/layers.py
ojustino: This seems like a good call. It's switched in the next commit.

ojustino:
OK, I changed `wwt.set_current_time` in `core.py` to also pass an ISOT string to WWT. (I also made a corresponding change in `wwt_json_api.js` .) I also integrated the `pytz` package to help verify/convert time zones for values provided to that method and in the time column of a table layer. **These steps eliminated the issues I was having with DST, and the points from the table in the `time_test` files now pop up when they should in both Qt and Jupyter** (on Firefox).

If this looks close to implementation we'll use going forward, I can go ahead and update the docs (for both table and image layers) and update the change log.

I wasn't sure how to handle adding a 'Z' to ISOT strings when some of them may already have a 'Z' or a '-05:00' at the end for time zone specification, so I didn't yet do so. I'm not sure if things are now synchronized because both `wwt.set_current_time` and the time series layer are in UTC or because they're both in the user's local time. Is there a simple way to test this?

ojustino:
**I went ahead and added a method to `core.py` that returns the current time in the viewer** -- `wwt.get_current_time()`. I followed @jsub1's methodology from #206 to ensure it works in Jupyter.

After reinstalling the directory with `pip` (due to changes to `wwt.js`), it seems to work as intended.

-----

I also made more edits to `TableLayer` so it can now read time columns made from a) `astropy` `Time`, b) `datetime.datetime` (with or without time zone information), or c) UTC ISOT strings. It then turns the column entries into ISOT strings with a specified "+0:00" time zone (as suggested by @astrofrog). **This removes the discrepancies in time between the Qt and Jupyter versions and ensures we are always working in UTC.**

All that's left out is ISOT strings _with_ time zones (e.g. `"2014-06-30T22:00:34+03:00"`). This is because I use `astropy`'s `Time` to verify them, and `Time` seems to only accept ISOT strings with no time zone info or with a 'Z' for UTC at the end. Should we try to support them, too?

I also corrected some tests in the validation of `time_att` that weren't doing what I thought they were.

pkgw:
@ojustino OK, the CI is now getting a bit farther. It's now erroring out because it thinks that some of your test files are pytest tests and tries to run them as such, which doesn't work. If you're ready to remove your test files, doing so would help the CI make more progress.

ojustino:
@pkgw OK, I guess they'll still be there in the old commits. Did you get a chance to test out the time series functionality or `wwt.get_current_time`? If you or @astrofrog could, it would be helpful.

pkgw:
OK, I've been exploring this and am seeing some weird stuff that I think is a lot more about how the WWT engine works than the pywwt layer. For reference, here's the outline of a test notebook I'm using:

```python
# ---- new cell ----
from pywwt.jupyter import WWTJupyterWidget
import numpy as np
from astropy import units as u
from astropy.table import Table
from astropy.time import Time

# ---- new cell ----
wwt = WWTJupyterWidget()
wwt.pause_time()
t0 = wwt.get_current_time()
wwt

# ---- new cell ----
n = 30
TIMESPAN = 100
ra_deg = np.linspace(0, 40, n)
dec_deg = np.zeros(n)
unix_times = t0.unix + np.linspace(0, TIMESPAN, n)
iso_times = [Time(ut, format='unix').isot for ut in unix_times]
table = Table({'ra': ra_deg, 'dec': dec_deg, 'time': iso_times})
layer = wwt.layers.add_data_layer(
    table,
    lon_att = 'ra', lat_att = 'dec', lon_unit = u.deg,
    time_att = 'time', time_series = True,
    size_scale = 60
)

# ---- new cell ----
wwt.set_current_time(t0)
wwt.play_time()
```

I'll see a dot show up, but in clumps instead of uniformly ... does WWT's time system lose precision at the few-second level? I'd be surprised.

Still trying to understand what's going on, but I wanted to write down the example that I'm looking at.

pywwt/core.py
```diff
@@ -220,14 +228,22 @@ def set_current_time(self, dt=None):
...
+            if dt.tzinfo is None:
+                utc_dt = pytz.utc.localize(dt).isoformat()
+            elif dt.tzinfo == pytz.UTC:
+                utc_dt = dt.isoformat()
+            else:  # has a non-UTC time zone
+                utc_dt = dt.astimezone(pytz.UTC).isoformat()
+        elif isinstance(dt, Time):
+            utc_dt = dt.to_datetime(pytz.UTC).isoformat()
+        else:
+            raise ValueError('Time must be a datetime or astropy.Time object')
```
pkgw: If I'm reading correctly, very similar code appears later; it should be centralized into a utility function.

pywwt/layers.py
```diff
@@ -54,10 +57,19 @@
...
+    NP_STR_TYPE = np.string_
+else:
+    STR_TYPE = str
+    NP_STR_TYPE = np.unicode_
+
 # The following are columns that we add dynamically and internally, so we need
 # to make sure they have unique names that won't clash with existing columns
 SIZE_COLUMN_NAME = str(uuid.uuid4())
 CMAP_COLUMN_NAME = str(uuid.uuid4())
+TIME_COLUMN_NAME = str(uuid.uuid4())
```
pkgw: This isn't specific to this PR, but shouldn't we hardcode these to constant values? Generating them on-the-fly just introduces a wildly obscure source of irreproducibility.

pywwt/layers.py
pkgw: As I think about it, I think I might actually prefer the idea of keeping the separate boolean. *If* we add code that guesses the time column automatically, it could be pretty confusing to add a layer and not see anything appear because it was automatically assumed to be time-dependent. On the other hand, overriding with a `time_att = None` wouldn't be so hard.

pywwt/layers.py
```diff
@@ -602,6 +682,59 @@ def _on_cmap_vmin_vmax_change(self, *value):
...
+        if isinstance(col, datetime):
+            wwt_times = Column([t.astimezone(timezone=pytz.UTC).isoformat()
+                                for t in col])
+        elif isinstance(col, Time):
+            wwt_times = Column([t.to_datetime(timezone=pytz.UTC).isot
+                                for t in col])
+        else:
+            col_list = col.tolist()
+            wwt_times = Column(col_list)
+        '''
```
pkgw: This string-comment stanza should be removed.

ojustino:
@pkgw I went over the results of `pytest` and fixed some that were failing due to changes I made in this PR. There are still 6 of 48 that fail -- a few mismatched images and a few that don't touch on time series layers.

I also updated the change log and moved the entry under 0.6.2 to 0.7.0 since it looks like we're going to skip the former.

ojustino:
The same behavior happens for me with your example, but if you try `time_test.ipynb` from commit ffa8d5b, the points appear one-by-one even though most are within a few (<5) seconds of one another. One difference is that my example is in solar system mode, though I don't know if that should matter.

ojustino:
Actually, your test now works as intended for me. Did you use `pip install -e .` after pulling the most recent changes to the PR? I tested on a different machine and that's what fixed things. (I could tell something was wrong because `wwt.get_current_time()` wasn't changing, and it only works after a fresh install since it involves some changed code in `wwt.js`.)

ojustino:
This was an arduous rebase; I have to do it over.

pkgw:
OK, I was able to track down my problem. It is in fact a precision issue in the WebGL shader implementation in the WWT SDK; the date calculations in SpreadSheetLayer are done relative to a `baseDate` field that is hardcoded to be January 1, 2010. Date differences that exceed floating-point precision away from that date will expose roundoff errors. So, if I change my example to fix it dates to be near `baseDate`, everything works OK:

```python
# ---- new cell ----
from pywwt.jupyter import WWTJupyterWidget
import numpy as np
from astropy import units as u
from astropy.table import Table
from astropy.time import Time

t0 = Time('2010-01-01T12:00:00.000', format='isot')

# ---- new cell ----
wwt = WWTJupyterWidget()
wwt.pause_time()
wwt

# ---- new cell ----
n = 30
TIMESPAN = 100
ra_deg = np.linspace(0, 40, n)
dec_deg = np.zeros(n)
unix_times = t0.unix + np.linspace(0, TIMESPAN, n)
iso_times = [Time(ut, format='unix').isot for ut in unix_times]
table = Table({'ra': ra_deg, 'dec': dec_deg, 'time': iso_times})
layer = wwt.layers.add_data_layer(
    table,
    lon_att = 'ra', lat_att = 'dec', lon_unit = u.deg,
    time_att = 'time', time_series = True,
    size_scale = 60
)

# ---- new cell ----
wwt.set_current_time(t0)
wwt.play_time()
```

We could reduce these issues by wiring up something to change `baseDate` to be (e.g.) the mean date of the table, or something like that. While this isn't an issue I want to spend too much effort on, it did come up in literally the first manual example I tried, so it seems like something that should at *least* be documented. I'll ponder a bit more ... anyone else have thoughts?

astrojonathan:
The mean data would definitely be better than the lower date especially if we can remove outliers that are effectively null dates.

PR #189, opened 2019-03-14
Installation and jupyter doc tweaks

ojustino:
I added minimum-version information about Jupyter-affiliated packages from #149 (hopefully resolving #80 as well) to the installation doc.

I also changed the image in the jupyter doc since it assigned `WWTJupyterWidget()` to `widget` while the text assumes it was assigned to `wwt`. The standard has been to use `wwt`, so I changed the image to match.


ojustino:
The quality of the image is sub-optimal, so I'll work on changing that...

ojustino:
OK, this should be ready to go once reviewed.

pkgw:
Can you keep the image file as a JPG rather than PNG? The difference in file size is nontrivial.

If you're comfortable rewriting Git history, it would be good to rewrite it to not add the PNG file and force-push to this branch. Otherwise the addition of the PNG file will add 1.5 MB to the repository size that will hang around forever.

ojustino:
@pkgw OK, the PNG file is erased from the history of this branch.

PR #192, opened 2019-03-28
Fix pywwt on `notebook >=5.7.6`

### LINKED ISSUE #191: `notebook` 5.7.6 breaks pywwt
[As described here](https://github.com/jupyter/notebook/issues/4467#issuecomment-471353499), version 5.7.6 of the Jupyter `notebook` module adds a new security feature related to the HTTP content types. This feature breaks pywwt.

On previous versions, the internal Jupyter server serves the `wwt_json_api.js` file with the wrong content/MIME type: it bills it as `text/html` rather than `application/javascript`. This is only a warning, though — in the Firefox JS developer console, you see:

> The script from “http://localhost:8888/wwt/wwt_json_api.js” was loaded even though its MIME type (“text/html”) is not a valid JavaScript MIME type.

The new `X-Content-Type-Options: nosniff` header added in `notebook` 5.7.6 turns this into a hard error:

> The resource from "http://localhost:8888/wwt/wwt_json_api.js" was blocked due to MIME type ("text/html") mismatch

This breaks programmatic control of the widget, basically.

This problem may be fixed in jupyter/notebook#4468? The PR there bills it as only applying to Windows, but looking at the patch, it might help pywwt out too.

pkgw:
Version 5.7.6 of the Jupyter notebook module adds a new security feature related to the HTTP content types (https://github.com/jupyter/notebook/issues/4467#issuecomment-471353499). It breaks pywwt because pywwt has not been returning valid MIME types (AKA content types) from its internal Jupyter server.

From looking at the main `notebook` implementation, it seems that the correct approach is to use the `mimetypes` module to get the right value to put in the Content-Type header, so let's go ahead and do that.

Closes #191.

### DIFF
--- a/pywwt/jupyter_server.py
+++ b/pywwt/jupyter_server.py
@@ -1,5 +1,6 @@
 import os
 import json
+import mimetypes
 from hashlib import md5
 from tornado import web
 from notebook.utils import url_path_join
@@ -32,6 +33,9 @@ class WWTFileHandler(IPythonHandler):
                 path = config['paths'][filename]
             else:
                 raise web.HTTPError(404)
+
+        # Do our best to set an appropriate Content-Type.
+        self.set_header('Content-Type', mimetypes.guess_type(filename)[0])
 
         with open(path, 'rb') as f:
             content = f.read()


pkgw:
The CI is failing here in the same way as #149. Importantly, failure mode is actually not the same as what I showed in [my first related comment on that issue](https://github.com/WorldWideTelescope/pywwt/pull/149#issuecomment-459729939). Here, and for the latest CI runs on #149, the "actual" image is all black. I can't reproduce on my Linux machine locally, so I'm not sure what's gone wrong ... any thoughts @astrofrog?

astrofrog:
@pkgw  - can you rebase to include the latest fixes to the CI? (https://github.com/WorldWideTelescope/pywwt/pull/196)

PR #195, opened 2019-04-04
Added validation for frames in TableLayer

ojustino:
Settling a to-do item in `layers.py`.

### DIFF
--- a/pywwt/annotation.py
+++ b/pywwt/annotation.py
@@ -13,6 +13,18 @@ from .traits import (Color, ColorWithOpacity, Bool,
 
 __all__ = ['Annotation', 'Circle', 'Polygon', 'Line', 'FieldOfView']
 
+def validate_traits(cls, traits):
+    '''
+    Helper function to ensure user-provided trait names match those of the
+    class they're being used to instantiate
+    '''
+    mismatch = [key for key in traits if key not in cls.trait_names()]
+    if mismatch:
+        raise KeyError('Key{0} {1} do{2}n\'t match any layer trait name'
+                       .format('s' if len(mismatch) > 1 else '',
+                               mismatch,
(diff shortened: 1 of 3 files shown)


ojustino:
I added another commit that explicitly lets the user know which of the keys for their calls to `TableLayer`, `ImageLayer`, or `Annotation` cause a `KeyError` for being invalid. Before, the messages were more vague in saying that "a key" was not valid, even if there were multiple offenders.

astrofrog:
@ojustino - can you rebase to include the latest fixes to the CI? (https://github.com/WorldWideTelescope/pywwt/pull/196)

ojustino:
@astrofrog I'm getting back on track and rebased this branch.

astrofrog:
@ojustino - sorry for not getting back to this sooner! Two requests before merging this:

* Could you make a helper function that basically does:

```
        mismatch = [key for key in kwargs if key not in self.trait_names()]
        if mismatch:
            raise KeyError('Key{0} {1} do{2}n\'t match any layer trait name'
                           .format('s' if len(mismatch) > 1 else '',
                                   mismatch,
                                   '' if len(mismatch) > 1 else 'es'))
```

since this block of code appears three times? You could make the function take a class and the kwargs and it could validate against the trait names of that class.

* Could you rebase this PR to include the latest improvements to the Continuous Integration?

ojustino:
@astrofrog Rebased. I added helper functions like you asked in both files. Did you mean for the function to appear once per file? The three instances are spread across two files, so in defining the function twice we only remove one instance of that block of code.

astrofrog:
@ojustino - can you define the function once in ``pywwt/utils.py`` and import it from there into both files?

ojustino:
@astrofrog OK, the code is now condensed.

PR #196, opened 2019-04-05
Fix CI issues

astrofrog:
This is just an experimental PR to try different things to fix the CI

### DIFF
--- a/pywwt/conftest.py
+++ b/pywwt/conftest.py
@@ -97,6 +97,13 @@ def wwt_qt_client():
     return wwt
 
 
+@pytest.fixture(scope='function')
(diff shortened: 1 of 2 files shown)


pkgw:
No dice, it looks like ...

astrofrog:
Ok so I haven't yet figured out a proper fix, but the summary of the issue is that somehow, after the loadFits call, the widget isn't quite reset to its original state, and changes to e.g. the background/foreground or the coordinates don't actually have any effect, so the images in ``test_full`` are just with the black background set in the image layer tests. For now I'll fix the CI by simply running the Image tests with an isolated Qt widget.

Issue #200, opened 2019-04-23
Proof-of-concept automagical interactive figure bundle output


pkgw:
The almost-finished [AAS lightcurve tool](https://aas-timeseries.readthedocs.io/en/latest/) is a framework for creating interactive plots of time-series data ([example](https://aas-timeseries.readthedocs.io/en/latest/views.html)). It has a really neat feature where you can create a figure interactively in a [Jupyter notebook](https://jupyter.org/), then use [one function call](https://aas-timeseries.readthedocs.io/en/latest/api/aas_timeseries.InteractiveTimeSeriesFigure.html#aas_timeseries.InteractiveTimeSeriesFigure.export_interactive_bundle) to export your creation to a Zip file in the format needed to embed it as an [interactive figure](https://journals.aas.org/graphics-guide/#interactive_figures) in a research article in an AAS journal. So the framework not only empowers astronomers to do cool, interactive data analyses, but it makes it really easy to memorialize them it in real-deal research publications. That's really cool!

pywwt should provide similar functionality! We already have the interactive Jupyter widget down — we just need to build the infrastructure to serialize its state to disk, then to wrap it all in a nice function call such as the `export_interactive_bundle()` referenced above.

It will probably be very difficult to serialize *all* possible pywwt operations to a standalone HTML file, which is why this issue is titled "proof-of-concept" — if we can cover the bases of serializing the view, basic tabular data, and basic images, that will still be a very powerful functionality that researchers will find very exciting.

pkgw:
Closed by #215.

PR #206, opened 2019-05-10
Fix get_center method for the jupyter widget and add a get_fov method for both widgets

jsub1:
Part of the work needed to implement issue #200. This uses traitlets to connect the jupyter widget's properties back into the python class.

### DIFF
--- a/lib/wwt.js
+++ b/lib/wwt.js
@@ -12,6 +12,10 @@ var WWTModel = widgets.DOMWidgetModel.extend({
         _view_module : 'pywwt',
         _view_module_version : '0.6.0',
 
+        _ra : 0.0,
+        _dec : 0.0,
+        _fov : 60.0,
+
     })
 });
 
@@ -35,6 +39,18 @@ var WWTView = widgets.DOMWidgetView.extend({
 
         WWTView.__super__.initialize.apply(this, arguments);
 
+        var self = this;
+        setInterval(function () { self.update_view_data(); }, 100);
+    },
+
+    _try_init_wwt_window : function() {
+        iframe = this.el.getElementsByTagName('iframe')[0];
+        if (iframe != null) {
+            this.wwt_window = iframe.contentWindow;
+            return true;
+        } else {
+            return false; // Not ready yet
+        }
     },
 
     render: function() {
@@ -106,13 +122,8 @@ var WWTView = widgets.DOMWidgetView.extend({
 
     handle_custom_message: function(msg) {
 
-      if (this.wwt_window == null) {
-        iframe = this.el.getElementsByTagName('iframe')[0];
-        if (iframe != null) {
-          this.wwt_window = iframe.contentWindow;
-        } else {
-          return  // Not ready yet
-        }
+      if (this.wwt_window == null && !this._try_init_wwt_window()) {
+        return;
       }
 
       if (msg['url'] != null && msg['url'].slice(4) == '/wwt') {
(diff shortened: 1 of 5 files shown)


lib/wwt.js
```diff
@@ -35,6 +39,18 @@ var WWTView = widgets.DOMWidgetView.extend({
 
         WWTView.__super__.initialize.apply(this, arguments);
 
+        var self = this;
+        setInterval(function () { self.update_view_data(); }, 100);
```
pkgw: If I'm reading this correctly, we set up to call `update_view_data()` periodically every 100 ms, right? That's not the end of the world but it's a bit less than ideal — is this definitely necessary? If so, a comment in the code explaining why would be helpful.

pywwt/jupyter.py
```diff
@@ -53,6 +56,16 @@ def _send_msg(self, **kwargs):
     def _serve_file(self, filename, extension=''):
         return serve_file(filename, extension=extension)
 
+    def _get_view_data(self, field):
+        if (field == 'ra'):
```
pkgw: You can (read: should) leave out the parentheses here.

pkgw:
Also, the Travis CI failures are due to the "link check" stage of the docs build: [see log here](https://travis-ci.org/WorldWideTelescope/pywwt/jobs/530790254#L1798). That breakage is unrelated to this PR, but if you can update the relevant URLs while you're at it, that would be good to straighten out.

jsub1:
I couldn't figure out a way to do it so that the properties are only updated when the python code requests it. Specifically, I can't figure out a way to block the python code until the properties have been update or found to not need updating on the javascript side without just guessing how long that should/could take and sleeping for that much. If we can work that out, we can remove the periodic calls to update_view_data.

pkgw:
OK, mini literature review on the issue of getting data out of the JavaScript. It seems that this is a known issue:

- [StackOverflow thread from 2017](https://stackoverflow.com/questions/30902898/ipython-notebook-javascript-retrieve-content-from-javascript-variables), no super compelling solution
- [Open notebook issue from 2018](https://github.com/jupyter/notebook/issues/3533), points out existence of [Comm](https://jupyter-notebook.readthedocs.io/en/stable/comms.html) framework, still doesn't seem to solve the blocking issue though
- [Asynchronous Widgets docs from ipywidgets](https://ipywidgets.readthedocs.io/en/stable/examples/Widget%20Asynchronous.html)
- [ipython_blocking] package

[ipython_blocking]: https://github.com/kafonek/ipython_blocking

I *think* that I have a proof-of-concept based on the technique in [ipython_blocking] that works! Basically, you can block the Python code by manually running the Jupyter event queue and storing events until you see one indicating that your operation has completed. I haven't done a detailed implementation, but the basic approach seems functional.

@astrofrog what do you think? The timeout approach clearly has its weaknesses, but it ought to work pretty well. I certainly am a little wary of manually futzing with Jupyter's event loop but it *seems* to work ... and I can certainly imagine that other cases might crop up where we'll need the Python code to talk to the JS code, so a more general framework for doing so might be valuable.

pkgw:
I rebased this on `wwt/master` to hopefully get the CI passing (cf #209).

lib/wwt.js
pkgw: As per offline discussion with @astrofrog, we're OK with this approach for now. #210 will track the broader issue and potential solutions.

PR #207, opened 2019-05-13
Fix Table Layer issue in Windows

jsub1:
Windows would not display table layers due to the way we were previously normalizing line endings for WWT. This PR fixes that problem.

### DIFF
--- a/pywwt/layers.py
+++ b/pywwt/layers.py
@@ -103,7 +103,7 @@ def csv_table_win_newline(table):
     table.write(s, format='ascii.basic', delimiter=',', comment=False)
     s.seek(0)
     #Replace single \r or \n characters with \r\n
-    return re.sub(r"(?<=[^\r\n])(\r|\n)(?=[^\r\n])", "\r\n", s.read())
+    return re.sub(r"(?<![\r\n])(\r|\n)(?![\r\n])", "\r\n", s.read())
 
 
 class LayerManager(object):
--- a/pywwt/layers.py
+++ b/pywwt/layers.py
@@ -2,6 +2,8 @@ import sys
 import uuid
 import tempfile
 
+import re
+
 if sys.version_info[0] == 2:  # noqa
     from io import BytesIO as StringIO
 else:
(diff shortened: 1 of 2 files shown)


pywwt/layers.py
```diff
@@ -93,6 +95,16 @@ def pick_unit_if_available(unit, valid_units):
             return valid_unit
     return unit
 
+def csv_table_win_newline(table):
+    '''
+    Helper funcion to get Astropy tables as ASCII CSV with Windows line endings
```
pkgw: Typo "funcion".

pkgw:
Looks good to me. @astrofrog will you have a chance to quickly look at this in the near future?

pywwt/layers.py
```diff
@@ -483,13 +495,7 @@ def _table_b64(self):
...
         # WWT absolutely needs that upon creation.
 
-        s = StringIO()
-        self.table.write(s, format='ascii.basic', delimiter=',', comment=False)
-        s.seek(0)
-
-        # Enforce Windows line endings
-        # TODO: check if this needs to be different on Windows
-        csv = s.read().replace('\n', '\r\n')
+        csv = csv_table_win_newline(self.table)
```
astrofrog: Does the behavior need to be different on Windows vs other platforms? I'm not sure why the code was the way it was before, but it seems to me that we only need to fix newlines when on Windows? (in which case we could use the ``platform`` module to check).

jsub1: WWT requires CSV data to use Windows line endings, but Astropy writes using the native line endings. The code that was there was designed to convert Mac line endings to Windows, but it messed up the line endings on Windows.

astrofrog:
I understand this fix better now, thanks! I think a cleaner fix would be to do:

```python
s = StringIO(newline='\r\n')
t = Table()
t['a'] = [1, 2, 3]
t['b'] = [3, 4, 5]
t.write(s, format='ascii.basic', delimiter=',', comment=False)
```

However this doesn't seem to work properly for now due I think to issues in Astropy.

Before I approve this, could you add a simple test for ``csv_table_win_newline`` as this will allow us to make sure this continues to work on all platforms in future?

jsub1:
@astrofrog I just pushed a quick test for the line endings.

pkgw:
OK, the CI failure is an unrelated, transient issue. We pass everywhere else so I think we're good to go here!

PR #208, opened 2019-05-15
Add FITS MIME type

jsub1:
The python mimetype library does not include the MIME type information for FITS files by default, so we have to add it. Otherwise adding FITS images breaks pywwt on notebooks 5.7.6 and up. (See #191 and #192 for more info on notebooks and MIME types)


pkgw:
I rebased this on the latest `wwt/master`, which should fix the CI issue related to checking links in the docs.

PR #212, opened 2019-05-29
Fix widget reset bugs

jsub1:
Fixes a couple of issues I found in the reset method:
1. Only half of the layers were removed when resetting the widget.
2. An exception was thrown when resetting the jupyter widget because we were also resetting the inherited DOMWidget traits.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ env:
   global:
     - SETUP_XVFB=True
     - CONDA_CHANNELS="astrofrog/label/dev"
-    - CONDA_DEPENDENCIES="astropy qtpy traitlets ipywidgets>=7.0 ipyevents widgetsnbextension pyqt pytest pytest-cov pytest-remotedata>=0.3.1 requests nomkl matplotlib beautifulsoup4 lxml jupyterlab flask flask-cors pyopengl"
+    - CONDA_DEPENDENCIES="astropy qtpy traitlets ipywidgets>=7.0 ipyevents widgetsnbextension pyqt pytest pytest-cov>=2.6.1 pytest-remotedata>=0.3.1 requests nomkl matplotlib beautifulsoup4 lxml jupyterlab flask flask-cors pyopengl"
     - PIP_DEPENDENCIES="sphinx-automodapi numpydoc sphinx_rtd_theme pytest-faulthandler codecov reproject"
   matrix:
     - PYTHON_VERSION=2.7
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ notifications:
 env:
   global:
     - SETUP_XVFB=True
-    - CONDA_CHANNELS="astrofrog/label/dev"
+    - CONDA_CHANNELS="astrofrog/label/dev conda-forge"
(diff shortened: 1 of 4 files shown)


astrofrog:
The issues in the tests are real - I think the problem is that ``foreground`` and ``background`` aren't traits that have the ``wwt`` tag (deliberately). Is that change really needed? What traits are being incorrectly reset?

jsub1:
There are about 10 different traits inherited from the DOM widget or one of its ancestors where trying to restore them to a default value will cause an exception to be thrown (`_dom_classes`, `comm`, etc.).

We could iterate over and reset the traits that are specifically defined in the base widget class, instead of all of the traits that a widget has.

astrofrog:
> We could iterate over and reset the traits that are specifically defined in the base widget class, instead of all of the traits that a widget has.

Yes, though we should make sure we are sub-class friendly too in case anyone ever defines traits in a subclass.

jsub1:
> > We could iterate over and reset the traits that are specifically defined in the base widget class, instead of all of the traits that a widget has.
> 
> Yes, though we should make sure we are sub-class friendly too in case anyone ever defines traits in a subclass.

The most general/transparent solution I can think of is to add a `wwt_reset` tag to the traits that we want to reset.
Another solution would be to loop over both `BaseWWTWidget.class_traits()` and `self.class_own_traits()`, but that wouldn't allow as much control as the above solution (and would break if we ever add a widget that doesn't inherit directly from the BaseWWTWidget).

astrofrog:
Adding ``wwt_reset`` works for me!

pkgw:
That sounds like a good solution to me too!

AppVeyor CI failure is an issue of antialiasing and color intensity (opacity?) of overdrawn lines in the Python 3.6 build.

Travis CI failures are due to an error in `make -C docs linkcheck` with what looks like a transient issue with GitBook, or possibly a too-short timeout because the problematic URL (<https://worldwidetelescope.gitbook.io/layer-control-reference/lcapicommands#table-of-properties>) does take a really long time to load.

@astrofrog OK to go ahead and merge?

jsub1:
One of the Travis failure (3.5) is due to the incompatibility of pytest 4.5 and pytest-remotedata 0.3.0 (Relevant issue: https://github.com/astropy/pytest-remotedata/issues/37). Can we make sure it grabs remotedata 0.3.1 or later going forward?

pkgw:
Since I went to the trouble of generating them, here are the images from the AppVeyor CI failure.

### expected
expected

### actual
actual

pkgw:
@jsub1 Hmm, not sure why it's pulling in the older version, but let's try forcing >= 0.3.1 and see what happens ... I just added a commit to your branch doing so.

pkgw:
Hmm, on Linux/Py35 [it says that it's using remotedata 0.3.1](https://travis-ci.org/WorldWideTelescope/pywwt/jobs/540832130#L2200) but the errors are still happening.

pkgw:
Oh, looks like we need [pytest-cov >= 2.6.1 as well](https://github.com/pytest-dev/pytest/issues/4608#issuecomment-470476524). Commit pushed.

pkgw:
Aha! The Travis CI doesn't use conda-forge to source packages, while AppVeyor does. I think these various issues have been due to getting older versions of packages from the more conservative Anaconda `defaults` channel.

Since AppVeyor uses conda-forge and we've just added pywwt to conda-forge, I've tried adding it to the Travis configuration here. Hopefully this will make these issues go away. And in recent runs, the GitBook issue has disappeared, meaning that the only CI problem left should be the AppVeyor/Py36 issue. I hope.

pkgw:
OK! The only Travis issue was a transient link-checking issue, so we've at least made progress on the CI front.

@astrofrog look OK?

PR #215, opened 2019-06-05
Interactive Figure Proof of Concept

jsub1:
Initial implementation of #200. This allows the user to save most of the current state of their WWT widget by calling the save_as_html_bundle function on the widget and specifying a folder or zip file destination. There is still functionality that would be good to add to the feature, some of which is listed below:

* Allowing users to name the exported csv and fits files with a nice interface (this pull request names the data files with the layer id)
* Better handling of errors when loading data:
    * Retry loading depending on returned error.
    * Better formatted way to display error messages.
* Add support for camera rotation and angle (and potentially add the ability to change these in python code)
* Potentially adding the option for the user to embed table data in the configuration JSON file
* Potentially show loading message and/or hide render window in the exported html page until all of the data is loaded.

### DIFF
--- a/pywwt/annotation.py
+++ b/pywwt/annotation.py
@@ -66,14 +66,14 @@ class Annotation(HasTraits):
     def _serialize_state(self):
         state = {'shape': self.shape,
                  'id': self.id,
-                 'settings': []}
+                 'settings': {}}
         for trait in self.traits().values():
             wwt_name = trait.metadata.get('wwt')
             if wwt_name:
                 trait_val = trait.get(self)
                 if isinstance(trait_val, u.Quantity):
                     trait_val = trait_val.value
-                state['settings'].append({'name': wwt_name, 'value': trait_val})
+                state['settings'][wwt_name] = trait_val
         return state
 
     def remove(self):
@@ -161,8 +161,7 @@ class Circle(Annotation):
 
     def _serialize_state(self):
         state = super(Circle, self)._serialize_state()
-        state['settings'].append({'name': 'skyRelative',
-                                  'value': self.radius.unit.is_equivalent(u.degree)})
+        state['settings']['skyRelative'] =  self.radius.unit.is_equivalent(u.degree)
         state['center'] = {'ra': self._center.ra.deg,
                            'dec': self._center.dec.deg}
         return state
--- a/pywwt/annotation.py
+++ b/pywwt/annotation.py
@@ -48,6 +48,7 @@ class Annotation(HasTraits):
             super(Annotation, self).__init__(**kwargs)
         else:
             raise KeyError('a key doesn\'t match any annotation trait name')
+        self.parent._annotation_set.add(self)
 
     def _on_trait_change(self, changed):
         # This method gets called anytime a trait gets changed. Since this class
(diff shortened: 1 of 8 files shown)


jsub1:
@pkgw I just rebased this onto the master branch.

jsub1:
> Initial implementation of #200. This allows the user to save most of the current state of their WWT widget by calling the save_as_html_bundle function on the widget and specifying a folder or zip file destination. There is still functionality that would be good to add to the feature, some of which is listed below:
> 
> * Allowing users to name the exported csv and fits files with a nice interface (this pull request names the data files with the layer id)
> * Better handling of errors when loading data:
>   
>   * Retry loading depending on returned error.
>   * Better formatted way to display error messages.
> * Add support for camera rotation and angle (and potentially add the ability to change these in python code)
> * Potentially adding the option for the user to embed table data in the configuration JSON file
> * Potentially show loading message and/or hide render window in the exported html page until all of the data is loaded.

Forgot to add that support for user added image collections is currently missing

pywwt/nbextension/static/interactive_figure/index.html
```diff
@@ -0,0 +1,29 @@
...
+    <script src="scripts/interactive_figure.js"></script>
+
+    <!-- The following is to avoid scrollbars -->
+    <style type="text/css">
+      body {overflow:hidden;}
+    </style>
+
+    <script>
+        $(document).ready(setHmtlSettings);
+        $(window).resize(setHmtlSettings);
```
pkgw: Typo: `Hmtl`, which propagates around a few places.

pywwt/nbextension/static/interactive_figure/index.html
```diff
@@ -0,0 +1,29 @@
...
+    </script>
+
+  </head>
+
+  <body onload="initialize()" style="margin: 0; padding: 0">
+      <div id="WWTCanvas" style="width: 100; height: 100; border-style: none; border-width: 0px;"></div>
+      <div id="WWTErrorText"></div>
```
pkgw: I think I'd like to use `kebab-case` identifiers for HTML element ids ... that seems to be the modern web style. So, can we do `wwt-canvas` and `wwt-error-text`? (This is relatively important since I think we can expect people to copy/paste this code a lot.)

pywwt/nbextension/static/interactive_figure/interactive_figure.js
```diff
@@ -0,0 +1,288 @@
+var wwtIntialState;
```
pkgw: Typo `Intial`, also propagates around a few places.

pywwt/nbextension/static/interactive_figure/interactive_figure.js
```diff
@@ -0,0 +1,288 @@
...
+
+var wwt;
+var wwt_ready = 0;
```
pkgw: Inconsistent naming with `wwt_ready`; I think it's good to aim to consistently use `lowerCamelCase` for JS identifiers, in which case we'd have `wwtReady`.

jsub1: I ended up removing `wwt_ready` because it’s not useful for the interactive figure. (It was used to signal back to pywwt that the widget was ready).

pywwt/annotation.py
```diff
@@ -62,6 +63,26 @@ def _on_trait_change(self, changed):
                                   setting=wwt_name,
                                   value=changed['new'])
 
+    def _serialize_state(self):
+        state = {'shape': self.shape,
+                 'id': self.id,
+                 'settings': []}
```
pkgw: Here and for the layers, the settings are stored as a list of dicts, each of which is currently a name/value pair. Should we instead store the settings as a dictionary instead?

```
"settings": {
  "opacity": 1.0
}
```

One obvious thing to think about is whether a setting will ever need to contain a complex value, but this can be accomplished by sometimes using a sub-dictionary:

```
"settings": {
  "opacity": 1.0,
  "color": { "r": 255, "g": 100, "b": 123, "my-metadata": "added on 2019 June 25" }
}
```

This change would affect the data storage format so I'd like to make sooner rather than alter if that's the path we go down. It's not a huge difference, but if you imagine altering the JSON structures with code, I think it will make life substantially easier to have a dict interface rather than iterating through lists (as in the test module right now).

pkgw:
Travis failures are transient link-check issues, so I think we're good!

PR #217, opened 2019-06-05
Work on image-display user experience

pkgw:
(This PR is cumulative on #216)

The main goal here is to add controls to the Jupyter widget so that the user can interactively tune the image stretch parameters straight in their notebook.

Screenshot from 2019-06-05 14-34-46

I set this up so that the Qt version of the widget should be unaffected.

I also ran into some issues with hidden JupyterLab widget views that look pretty thorny; I have some lengthy commit messages describing what I found and a workaround I put into place.

### DIFF
--- a/lib/wwt.js
+++ b/lib/wwt.js
@@ -135,10 +135,30 @@ var WWTView = widgets.DOMWidgetView.extend({
         msg['url'] = this.wwt_base_url + msg['url'];
       }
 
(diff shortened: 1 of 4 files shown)


pkgw:
This needs docs, but I'm going to merge now. (I'm also going to ignore the minor hit in code coverage.) I want to rearrange the docs a bit anyway, so that will wait til another pull request.

PR #218, opened 2019-06-06
Try to get Windows CI working again

pkgw:
When I added the hack to get Python 3.6 running on our AppVeyor Windows CI, I think everything passed, but now we get different images in our testing framework. At the moment, the version-constrained CI uses WebKit as its rendering engine instead of WebEngine.

The hack was to fix the version of pyqt to avoid a library incompatibility issue, so let's try applying that constraint *everywhere*, which will hopefully get all of the Windows tests using WebKit. Then we can modify the test images to match the differences between the two engines.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,15 +13,14 @@ environment:
                         # to the matrix section.
 
       CONDA_CHANNELS: "astrofrog/label/dev conda-forge"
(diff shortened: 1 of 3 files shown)


pkgw:
Huzzah! Note that the approach taken in the merged PR is a bit different than that outlined in the PR description; see the commit messages for up-to-date info.

PR #219, opened 2019-06-06
pywwt/jupyter_server.py: fix file service in JupyterHub contexts

pkgw:
When we're serving up files in JupyterHub, we need to include the notebook server's base_url in the URL we generate. This is obnoxious to determine from inside a kernel, but I found some code that shows how to do it.

Also added some module-level documentation outlining what's going on and why.


pkgw:
Travis failures are transient issues with link checking, so I'm going to go ahead and merge.

PR #239, opened 2019-09-25
setup.py: fix package build on BinderHub

pkgw:
For whatever reason, the package build reads in README.rst such that it expects it to be ASCII encoded, leading to an error if there's any Unicode. Use io.open and specify the encoding manually to make sure this doesn't happen.


pkgw:
CI failures are spurious linkcheck failures and this is blocking the BinderHub update, so I'm going to go ahead and merge despite the failures.

PR #240, opened 2019-09-25
Changed Binder links on index page of docs

ojustino:
Here are my proposed changes to the Binder links on our docs' index page.


pkgw:
CI failure are once again only due to a transient-looking problem with `requests` in the docs linkcheck stage, so, merging.

PR #259, opened 2020-03-30
Infrastructure cleanup

astrofrog:
This makes the following improvements:

* Update some SDK API calls that have changed
* Fix some of the image tests that only run locally (not in CI)
* Add tox configuration
* Fix code style issues
* Switch from flask to tornado to make data server quieter
* Exclude astropy 4.0.1 from dependencies since it breaks reproject 0.6
* Simplify CI set-up to use OpenAstronomy Azure template (https://openastronomy-azure-pipelines.readthedocs.io/en/latest/)
* Drop Python 2.7 support

The Windows builds have intermittent segfaults but this is seen in master too, so no change there.


pkgw:
Thanks for working on all of this stuff, Tom!

astrofrog:
@pkgw - this is ready for review/merging - the Windows jobs fail intermittently but that also happens on master and is unrelated. For now I've left PyQt5 pinned to 5.9, but I think there may be some issues with the latest releases (to be investigated separately)

pkgw:
@astrofrog What issues arise with the latest Qt? I believe that we need Qt >= 5.12 on Linux due to some SSL-related shenanigans.

And, sigh, it looks like the Linux test has hung.

astrofrog:
Actually Qt 5.9 seems to work fine on Linux. With Qt 5.14 I see WebGL rendering issues - I can open a separate issue. Not sure why the linux job hung, weird.

PR #261, opened 2020-04-03
Preparing release v0.8.0

astrofrog:
I'm preparing the v0.8.0 release and at the same time trying to switch to use the OpenAstronomy Azure Pipelines template for auto-releasing when a tag is pushed. I tried to keep the release instructions as close to the original as possible. In practice, for other packages I don't bother with the release branch and PR, so the process is simpler (update changelog and versions, push, then once CI passes, tag, push to master).


pkgw:
Awesome, thanks!

lib/version.js
```diff
@@ -0,0 +1 @@
+export const version = require('../package.json').version;
```
pkgw, line 1: Is this going to work in the created/exported package? I.e. does it always include the `../package.json` file?

astrofrog: It should - we are doing it already for another package (bqplot-image-gl).

Issue #273, opened 2020-09-23
Some imagery layers don't work


astrofrog:
The following example doesn't work:

```
from pywwt.qt import WWTQtClient
%gui qt
wwt = WWTQtClient()
wwt.foreground = wwt.imagery.other.planck2
```

(``wwt.imagery.other.planck2 == 'Planck Spinning Dust'``)

@pkgw - just to check, any ideas why this might be happening?

pkgw:
Ah, right, this issue. So the problem is that the WebGL engine has a somewhat annoying thing where it has an internal list of available image sets loadable via (e.g.) `loadForegroundImageByName`. When you load a new WTML file, its imagesets do **not** necessarily get added to that list. By default the WebGL engine can only load-by-name the imagesets listed in this file:

http://web.wwtassets.org/engine/assets/builtin-image-sets.wtml

That file contains a variety of data sets, including a couple of Planck ones (in terms of sheer numbers, it's dominated by panorama data sets), but not the spinning dust.

Meanwhile the list of surveys in pywwt is derived from:

https://worldwidetelescope.github.io/pywwt/surveys.xml

(which in turn derives from http://www.worldwidetelescope.org/wwtweb/catalog.aspx?W=surveys)

which is the "All-Sky Surveys" folder of the default webclient explore view. In pywwt we feed `surveys.xml` into `wwt.loadImageCollection` but it turns out that this does *not* cause the imagesets to be added to the table of nameable imagesets. So then `loadForegroundImageByName` and friends don't work with them.

Possible courses of action:

- Synchronize the pywwt list of images with `builtin-image-sets.wtml`, which is what it should have been all along.
- Make it so that the imagesets in `surveys.xml` actually get registered with the engine so that they are findable.

As a side comment, the existence of `surveys.xml` at all is a bit of an artifact from the times when we had very little control over the WWT web presence. Nowadays we could host such data files on `web.wwtassets.org` or something similar instead of GitHub pages.

pkgw:
Sorry, that wasn't 100% correct.

[As per the docs][licdoc], the `loadImageCollection` call does add the imagesets in the `surveys.xml` file to the internal list ... but only in the top level of the folder. As you might recall, a while ago we changed something in pywwt to recurse into subfolders as well. "Planck Spinning Dust" is in a "Planck" subfolder, which is why it doesn't get loaded.

[...]

PR #274, opened 2020-10-01
Added version of surveys.xml without nested folders

astrofrog:
@pkgw as you suggested this nicely solves #273 - will we add a new file or replace the existing one? (which would instantly fix all PyWWT installations)


pkgw:
Thanks! I don't see any reason not to replace the existing file, I think. As I sort of indicated in the earlier discussion, I'm interested in revisiting how this whole framework works, but we should fix the existing installations.

astrofrog:
Ok I’ll update this PR shortly to replace the file

astrofrog:
@pkgw - this should be good to go

PR #275, opened 2020-10-13
Isolate the JS package files in a subdirectory

pkgw:
### Overview

Isolate the `pywwt` NPM/JS package in a `frontend` subdirectory, rather than having it be intermingled somewhat confusingly with the Python package.

As part of that isolation, separate the `build` step, to generate a big bundled JS file with webpack, and what I'm calling a `pywwt-export` step, that copies the JS artifacts to their needed locations in the Python package tree. I think it would be more correct to have this "export" step be implemented fully on the Python package side, but given the existing infrastructure it's a bit simpler to implement it as an NPM script. 

### Checklist

- Will be tested in CI
- Documentation forthcoming after new build setup is worked out.


pkgw:
Update: ran into CI problems, initially triggered by API breakage in `traitlets > 5`, but also induced by trying to build on Pytho 3.8. I think that fundamentally, Tox is not a good testing tool for pywwt because it requires venvs for test environment setup, while we need to use Qt+OpenGL, which induce lots of platform-dependent gnarliness. So I am working towards using a conda-based test setup rather than Tox.

pkgw:
@astrofrog I'm seeing a bunch of failures in the pywwt image tests as I try to get this working (again) on Windows and Darwin. I've at least got the CI uploading the output test imagery as build artifacts, so that it's a bit easier to see what's going on. Do you have any memories of issues and/or debugging to get this stuff working? You had disabled macOS tests altogether, but Windows *was* working relatively recently ...

pkgw:
That took a while.

PR #276, opened 2020-10-17
Yet more CI robustness work

pkgw:
I thought that I could get away with impatience on a couple of macOS tests, but now I think not.


pkgw:
Xref: WorldWideTelescope/wwt-webgl-engine#67

PR #287, opened 2021-02-09
pywwt/jupyter.py: add support for the JupyterLab application

pkgw:
### Overview

Prototype-quality support for connecting pywwt to the freestanding WWT JupyterLab application, as opposed to individual "widgets" associated with specific notebooks.

In order to be able to share image data with this application, we need to make sure to set CORS data-sharing headers for the files that we serve up.

### Checklist

Docs and tests coming, but I want to make a release with the initial support to test out how everything hangs together in production.


pkgw:
CI failed spuriously. Reran in a new job (since our failing jobs uploaded their image debugging artifacts, meaning that subsequent runs would always fail with redundant artifacts) and it's all green.

PR #292, opened 2021-05-27
Companion patch to WebGL FITS engine release

imbasimba:
Slight changes for this PR
https://github.com/WorldWideTelescope/wwt-webgl-engine/pull/96/

### DIFF
--- a/pywwt/nbextension/static/interactive_figure/interactive_figure.js
+++ b/pywwt/nbextension/static/interactive_figure/interactive_figure.js
@@ -172,7 +172,17 @@ function loadImageLayer(layerInfo) {
                                stretchInfo['vmin'],
                                stretchInfo['vmax']);
         wwtLayer.set_colorMapperName(stretchInfo['cmap']);
-        layer.get_imageSet().get_fitsProperties().transparentBlack = false;
+
+        // old transparentBlack API, @wwtelescope/engine <= 7.10. With
+        // newer engines, this is a harmless no-op.
+        layer.getFitsImage().transparentBlack = false;
+        
+        // new transparentBlack API
+        var imageset = layer.get_imageSet();
+        if (typeof imageset['get_fitsProperties'] !== 'undefined') {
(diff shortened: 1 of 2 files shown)


pywwt/nbextension/static/interactive_figure/interactive_figure.js
```diff
@@ -172,7 +172,7 @@ function loadImageLayer(layerInfo) {
                                stretchInfo['vmin'],
                                stretchInfo['vmax']);
         wwtLayer.set_colorMapperName(stretchInfo['cmap']);
-        layer.getFitsImage().transparentBlack = false;
+        layer.get_imageSet().get_fitsProperties().transparentBlack = false;
```
pkgw, line 175: This kind of thing is why I'm always so concerned with API stability :-) In order to be able to make a new release of pywwt that won't break until the engine update is released, I think we need to simultaneously support both versions of the API. I think this would be the way to do it?

```js
// old transparentBlack API, @wwtelescope/engine <= 7.10. With
// newer engines, this is a harmless no-op.
layer.getFitsImage().transparentBlack = false;

// new transparentBlack API
var fp = layer.get_imageSet()['get_fitsProperties'];
if (typeof fp !== 'undefined') {
    fp.transparentBlack = false;
}
```

pkgw (commented):
Due to the way that pywwt is decoupled from the engine, I think that we need to support both versions of the API here.

imbasimba:
Yeah, you are completely right, of course. By now, you have imprinted the importance of this into my brain as well. It should be a long time before I make this kind of oversight again :)

pkgw:
Thanks! The CI issues here seem to be due to (1) a transient Conda failure and (2) a bug in the current Matplotlib Intersphinx file that results in the `linkcheck` action reporting a broken outgoing link that we can't do anything about. So, I'll merge this now. If we still need to avoid the Matplotlib issue by the time we need to make a release, we can just disable the `linkcheck` step for a little while.

PR #289, opened 2021-02-22
Catalog HiPS for notebooks

imbasimba:
core.py: DEFAULT_CATALOG_HIPS_URL needs to point to and xml of Catalog HiPS. 
Same structure as described here: https://github.com/WorldWideTelescope/wwt-webgl-engine/pull/84

PR #299, opened 2021-06-10
Some Jupyter/Lab/widget tidying

pkgw:
### Overview

Some tidying of the JS/Jupyter/Lab hairball as I think about how to integrate #289.

**Work-in-progress:**

- Rename `wwt.js` to `widget.js` in the frontend; that's more accurate
- Include @imbasimba's doc fix while I'm here
- Work on frontend docs/comments
- Remove long-neglected examples
- Remove some superfluous file reads in the Qt webserver infrastructure
- Move interactive_figure static files out of nbextension/static, since they don't need to be / belong there.
- Move webserver static files for the same reason.

### Checklist

N/A, no new functionality or bugfixes: this is refactoring.


pkgw:
@imbasimba I've been thinking about how to approach the Python/JS/HTML architecture for your #289, and as part of that I spent some time reminding myself how all of the moving parts fit together here (Python, WWT engine, Jupyter, JupyterLab, Qt ...). This PR will break yours, but I think it also sets us up to move away from the manually-coded HTML/JS here to start using the research app, which IMO would be a big win.

PR #302, opened 2021-07-19
Port the Qt client to use the research app

pkgw:
### Overview

Port the Qt client to the research app, making it so that all of our implementations are now using the same messaging API.

**DRAFT:**

- [x] Add no-chrome option to preserve current Qt look-and-feel
- [ ] Unify base widget implementations

### Checklist

- [x] Changed code has some test coverage (or justify why not)


pkgw:
This is very early work, but filing this now to get a sense of where we'll land with the test suite.

PR #308, opened 2021-07-23
Updated HiPS catalog support

pkgw:
### Overview

This PR supersedes #289, building on the improved foundation that I've been working on. Namely:

- Unifying all modes to to use the "app" as the viewer backend
- Here, adding foundational support for async operations

The async support turns out to require an extremely gross hack of the Jupyter kernel framework but I believe that the approach is fundamentally OK, if hacky and fragile.

### Checklist

- [ ] No test coverage :-(
- [ ] No docs :-(

### DIFF
--- a/frontend/lib/widget.js
+++ b/frontend/lib/widget.js
@@ -158,6 +158,7 @@ var WWTModel = widgets.DOMWidgetModel.extend({
             if (!this._kernelThinksWidgetIsAlive) {
                 this._kernelThinksWidgetIsAlive = true;
                 this.send({
+                    _pywwtExpedite: true,
                     type: 'wwt_jupyter_widget_status',
                     alive: true
                 });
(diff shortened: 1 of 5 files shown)


pkgw:
YOLO.

imbasimba:
Interesting hack of the kernel :)

To make it less fragile, we should probably control the allowed versions of the kernel a bit more. At least adding a minimum version for the ipykernel. It did not work with my previous ipykernel (5.5.4), but upgrading to the latest version (6.3.1) seems to solve the problem. So somewhere inbetween this hack starts to work, probably at 6.0.

pkgw:
@imbasimba Yes, that's a good idea. I'm hoping that this approach won't turn into a maintenance nightmare, but that depends a bit on how the ipykernel code evolves :-/

PR #311, opened 2021-08-12
User selections

Carifio24:
### Overview

This PR adds fields in the pywwt widget for observing changes in selection in the WWT app (the `SelectionStateMessage` proposed in https://github.com/WorldWideTelescope/wwt-webgl-engine/pull/133). Additionally, this adds functionality for a user client to add a callback to be called when the widget receives a selection message.

So this is what I was thinking for the basic callback setup. The callback function runs after the state is updated (and after any futures are resolved as well), and is passed two arguments: the widget itself, and a list of the names of properties that were updated. This gives the caller access to the updated state and the flexibility to do whatever they want with it, but minimizes the number of attached callbacks (I'm thinking that a user could attach one callback for each message type).

### Checklist


pkgw:
OK, some details might have to evolve pending discussion in WorldWideTelescope/wwt-webgl-engine#133, but I like the overall approach here. I think it makes a lot of sense to have the message-handling code primarily support futures, but also provide callback hooks when there's a demonstrated need in a downstream user that can't go full async.

Carifio24:
I've updated this to match the final version in https://github.com/WorldWideTelescope/wwt-webgl-engine/pull/133 - namely, I've removed the field for selected HiPS catalogs, as well as the JSON parsing of the source fields that was previously present. I've also added a `CallbackError` exception for errors that occur within a user callback.

It looks like there's now an issue with one of the Qt tests, though, so I need to figure out what the issue is there.

pywwt/core.py
```diff
@@ -374,8 +378,8 @@ def _on_app_message_received(self, payload):
         if callback:
             try:
                 callback(self, updated_fields)
-            except Exception as e: # TODO: Something better here
-                print("Error when calling selection updated callback:\n%s" % str(e))
+            except Exception as e:
+                raise CallbackError() from e
```
pkgw, line 384: Hmmm. The thing to worry about here is that this function can be called from weird async Python event handlers where any exceptions might just disappear into the ether. I think the least-bad thing to do here is emulate what's done in `qt.py`, and use `from .logger import logger` paired with `logger.exception` to report any callback errors.

pywwt/core.py
```diff
@@ -892,6 +951,26 @@ def available_hips_catalog_names(self):
...
+        The most recent source selected in the viewer.
+        """
+        return self._most_recent_source
+
+    _selected_sources = []
+
+    @property
+    def selected_sources(self):
+        """
+        A list of the selected sources, as dictionaries.
```
pkgw, line 970: The docstrings here need to give the reader a more specific idea of how the return values are actually structured. In this particular case I think that it would work well to point the reader to the research-app-messages API docs:

https://docs.worldwidetelescope.org/webgl-reference/latest/apiref/research-app-messages/interfaces/selections.source.html

These docs could have a bit more detail themselves (e.g. specifying the units for angles) but they'll help someone get an idea of the general shape of the data.

Carifio24: Good idea, I'll point the reader in that direction. The `research-app-message` docs do give the units (of radians) for the angles, but they could use more information about the `catalogData` and `zoomDeg` fields.

pkgw:
CI failure has to do with some differences of the image tests on Linux — might be spurious? I've told Azure to relaunch those tests to find out.

pkgw:
Hmm, looks like the errors are reproducible. You should be able to go to the artifacts of the CI run and download packages of data showing the images generated during the failing runs, along with diffs against the reference images. (If you can't download the files there might be a permissions problem. If so, let me know.) There seem to be some teeny-tiny changes, just enough to fail the test, but nothing obviously broken. Azure might have upgraded its Linux builders recently? In which case, the solution would be to add the new image variations to the test corpus.

Carifio24:
I was able to download the generated images and diffs from Azure. The four images that failed all look 'good' to me (in the sense that, visually, I can't see any real difference between them and the reference images in `tests/data`).

The latest release on the Azure Pipelines agent repository is marked as being from 27 days ago, which is one day after the last commit here on `pywwt`, so it is possible that this is due to an Azure upgrade.

pkgw:
OK. I think you'll have to update the test corpus. Can you take the images that are currently failing and add them to the repo with filenames like `pywwt/tests/data/refimg_qt_full_step?/e.png` ?

Within the `refimg_` directories, the sample PNG files can be named anything. I just went with `a`, `b`, `c`, etc., rather than trying to give informative names (e.g. something like `linux_llvpmpipe`) since it's possible that one image will work well for multiple platforms, and that the "correct" image for a given platform might even evolve over time (as we're seeing here).

pkgw (approved):
Code changes look great now, just need to update the test corpus.

PR #313, opened 2021-09-21
Update dep version requirements: research-app 0.7, reproject 0.8

pkgw:
### Overview

In preparation for a new release, bump some dependency version requirements.

### Checklist

N/A


pkgw:
@Carifio24 Here's a mystery for you. The Windows and macOS builds are now failing with, among other things, Qt tests where the imagery seems to be shifted upwards from where it should be. (There are other failures, but they seem to be more reasonable.) I can't reproduce locally with a build my Windows partition. Any ideas about why we might be getting this failure mode? I've been trying to see whether using true OpenGL vs. software rasterization makes a difference (that's one thing that distinguished the Windows and macOS Azure Pipelines builds), but that's its own can of worms ...

pkgw:
If I'm analyzing this correctly, the strange vertical offsets don't happen when using the 0.6 version of the research app, and they don't happen when using Qt 5.15 on Windows (vs. Conda's 5.12), and they don't happen when using real GL as opposed to the software renderer.

pkgw:
OK, I tried changing the CI scripts to use Qt 5.15 (by installing with pip), and on macOS the vertical offset is still there, so that's not helpful. (On Windows, other issues cropped up.)

pkgw:
When I temporarily changed the package to use version 0.6 of the research app, the problem was fixed. (That's Pipelines build 20210922.2).

pkgw:
On my Windows machine, if I resize the physical WWT window, the offset seems to be stable regardless of what I change, so it seems like the view center is really shifted, and it's probably not an issue of the output bitmap getting cropped oddly or something.

pkgw:
The angular offset is about 2.8 degrees, compared to the image height of 10 degrees.

pkgw:
I got it to the point where I could bisect on different versions of the engine on my Windows partition. Not too surprisingly, the culprit seems to be somewhere in WorldWideTelescope/wwt-webgl-engine#115, the tiled FITS pull request.

pkgw:
@imbasimba This pull request upgrades pywwt to use the latest version of the WWT research app and engine, which turns out to cause new failures in the CI test suite.

I believe that I've traced this down to the WebGL 1 fallback code: e.g., the pywwt test `image_layer_equ` passes with the latest engine when WebGL2 is enabled, but fails when only WebGL 1 is used with the image being drawn too far up (i.e., at too high of a  declination on the sky). The failures in the CI suite here are on the platforms where we have to use software GL rasterization that apparently triggers the WebGL 1 fallback.

Would you be able to look into this and see if you can confirm my analysis? This version of pywwt isn't using the tiled fits functionality, so it would be something with the WebGL1 approach to SkyImage FITS images.

imbasimba:
Hmm... I'll look into this

imbasimba:
Here is the little thing I messed up in the SkyImage/TangentTile class merge.
https://github.com/WorldWideTelescope/wwt-webgl-engine/pull/147

pkgw:
Thanks for investigating! My local testing looks like it confirms that you've found the problem.

pkgw:
@imbasimba This pull request now adds some documentation in the README with a bit more info about how the whole image-test subsystem works.

imbasimba:
I found the reason for a lot of the confusion and conflicting results and it is no wonder it was so difficult to diagnose and fix this issue! There were two separate problems interfering with my investigation. On top of the real bug there is a later commit slightly changing the zoom level https://github.com/WorldWideTelescope/wwt-webgl-engine/commit/aeb5dc042c5ebe2f330e582b139321e272d4374d causing the tests to fail when I was working on master.

This explains so much of my conflicting results! I'm glad I'm not crazy 😃

imbasimba:
Great, thanks for the added documentation!

pkgw:
Yeah, sorry that this is all such a mess :-/ From a certain standpoint, these image tests should be maintained in the wwt-webgl-engine repo to keep them "closer" to the engine changes.

Side note: as mentioned in one of the commit messages in this PR, I noticed that the WebGL1 and WebGL2 renderers give fairly different results when opacities are used, as in the `image_layer_gal` test. I was wondering if maybe the WebGL1 path was doubly-applying the opacity (its 50% opacity looks much fainter than WebGL2), but as far as I can tell they just handle the opacity in different ways that lead to different "response curves".

imbasimba:
From a quick look it seems like the main difference is that opacity scales rgba in TileShader while in FitsShader opacity only affects the alpha channel. I can do a quick test before our next meeting to see if that's the only reason for the discrepancy.

pkgw:
Yeah. I guess if opacity is scaling RGB *and* A, that would be incorrectly doubling its effect, right? Assuming that those colors aren't supposed to be premultiplied alpha.

But overall, it's not something to prioritize, I'd say.

imbasimba:
Indeed the RGB is scaled twice in all shaders except the FitsShader. Basically this [blender function](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/blendFunc) makes the rgb * alpha unnecessary. 
https://github.com/imbasimba/wwt-webgl-engine/blob/7a010aaf8d56d25a69ce05871d9a68a489b62d75/engine/wwtlib/Graphics/Shaders.cs#L1677:L1678

Of course one could argue that there is no single correct answer when it comes to merging two images, but this _seems_ like an honest mistake.

PR #316, opened 2021-11-03
Improved tiled FITS support

imbasimba:
First draft of Single command FITS tiling & additions. I'll continue to evolve this, especially regarding the async/await requirements, but early feedback is very welcome.

As you can see I based this PR on @pkgw's kernel-data-relay branch. There are some merge conflicts between the kernel-data-relay and the master branch which I have not delved into yet.


imbasimba:
I finally found a way to go around the async user experience dilemma. I.e. users have to remember if they have to write ‘await’ before the function call or not. I’m doing this with the help of the nest_asyncio library (https://github.com/erdewit/nest_asyncio), which, to my knowledge, is the only way to achieve the desired result in jupyter. This library enables nested event loops, which essentially allows us to start an async function from a sync wrapper function. Thus the user does not have to directly interact with the async functions. I do not see any potential problems in doing this, at least not if we only use nest_asyncio to wrap user-facing async functions. But maybe @pkgw with your in depth knowledge of the jupyter and async systems, you can see some potential problems with this approach?

As a bonus, this should allow us to just have one ‘add_image_layer’ type function :)

pkgw:
@imbasimba Cool! It seems that the `jupyter_client` package already includes some support for `nest_asyncio`, which I think is a good sign.

By the way, the merge conflict is probably because I ~"backported" a change to `jupyter_server.py` to the main branch in #315. That in turn enabled us to get the app working in [SciServer](https://www.sciserver.org/), which as of a few days ago is now deployed in their HEASARC 6.29 image! That server build also includes the Kernel Data Relay, so once this work makes it into a release, SciServer will be ready to make use of it!

pkgw:
I've released the new toasty functionality (version 0.13) and pushed an update to this branch that should start using the API under its published name.

pkgw:
For the record, the Azure Pipelines CI should be all green now — there was a transient failure, and then the current build is only marked as failing because the rerun job can't upload duplicate test-result artifacts.

pkgw:
OK. Taking another look over the code, I think this is ready to merge! I think that the changes here are all looking pretty much good, and the diff is pretty substantial due to my reformatting, so I'd rather tidy up any issues in another PR. And likewise I'd really like to make sure that docs for the new APIs are in good shape, but that's part of a much bigger issue where I'm just dissatisfied with the appearance of the pywwt API docs in general.

PR #317, opened 2021-11-04
setupbase.py: work around breakage in setuptools 58.5

pkgw:
### Overview

The most recent version of setuptools has a bug that breaks our
install: https://github.com/pypa/setuptools/issues/2849

Here we copy the solution from
https://github.com/scikit-image/scikit-image/pull/6007 .

### Checklist


pkgw:
Note, this PR also reformats the file using "black", which I've turned on as a default in my editor.

PR #319, opened 2021-11-29
More add layer options

imbasimba:
Adapted add_image_layer for new toasty functionality (including HiPSgen) https://github.com/WorldWideTelescope/toasty/pull/69

- hdu_index can now also be a list
- Previous usage of hdu_index with non-tiled FITS caused KeyError by ImageLayer's validate_traits. Now fixed by a kwargs.pop. 

@pkgw Can you think of a better way of dealing with the toasty keywords? Right now, they are manually popped in _remove_toasty_keywords, which doesn't feel too stable. At the same time validating the traits sent to ImageLayer seems like a useful thing to do. And exposing all toasty settings explicitly in the add_image_layer does not seem optimal either.  Let me know what you think.

### DIFF
--- a/pywwt/layers.py
+++ b/pywwt/layers.py
@@ -238,7 +238,15 @@ class LayerManager(object):
         self._layers = []
         self._parent = parent
 
-    def add_image_layer(self, image=None, hdu_index=None, verbose=True, force_hipsgen=False, force_tan=False, **kwargs):
+    def add_image_layer(
+        self,
+        image=None,
+        hdu_index=None,
+        verbose=True,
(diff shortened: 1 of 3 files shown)


pkgw:
I've force-pushed an update this branch to rebase it against `master` (which introduces some nice improvements in the docs styling) and tidy up a few things, including formalizing the dependency on the just-released toasty 0.14. I've "fixed" an issue dealing with images with non-celestial axes here, where `reproject` was just rejecting them. I've basically copy/pasted the analogous code in Toasty to drop the non-celestial axes.

PR #321, opened 2021-12-04
Increased jupyter-relay chunk size to 8MB

imbasimba:
I have not found the root cause of this issue, but increasing the chunk size to something ridiculously high, like 8MB should make this error extremely unlikely.

What I found during the investigation is that the messages are all sent correctly, but only the first 2000 are received by KDR.

In some rare instances, when multiple requests are fired almost simultaneously, less than 2000 are received. This has led me to believe that the limit is on the receiving end, so we cannot just make sure that every single request is less than 2000 messages, since all the requests are fighting for the same 2000 spots. So I felt like it is safest to up the chunk size to something very, very high. At least until the root issue is found and resolved.

### DIFF
--- a/pywwt/jupyter_relay.py
+++ b/pywwt/jupyter_relay.py
@@ -322,7 +322,7 @@ class JupyterRelayHub(object):
 
             while keep_going:
                 # The chunk size here is just a guess:
-                chunk = handle.read(32768)
+                chunk = handle.read(8388608)
 
                 if len(chunk):
                     buffers = [chunk]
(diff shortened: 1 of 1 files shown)


pkgw:
Frustrating that the root cause is hard to find here, but I just made up the initial chunk size so I think the approach here is the way to go for now. You just had a complaint from `flake8` about the code formatting here.

(On that note, I think that it would be nice to transition all of our Python codebases to be auto-formatted using [`black`](https://black.readthedocs.io/en/stable/). I've found that using the "format on save" mode in my editor (VS Code, but most other editors support it these days I think) is very helpful for achieving this, but I keep on turning the mode on and off since I work with a ton of legacy codebases for which I definitely do *not* want auto-formatting.)

PR #322, opened 2021-12-21
Removed pywwt.annotation.rst

imbasimba:
I'm not entirely sure why there are two pywwt.annotation.rst, but the are fighting each other since they have the same name with different capitalization (pywwt.**a**nnotation.rst vs pywwt.**A**nnotation.rst)

So I deleted one, but maybe there is a better solution?

### DIFF
--- a/docs/api/pywwt.annotation.rst
+++ b/docs/api/pywwt.annotation.rst
@@ -1,3 +1,0 @@
-.. automodapi:: pywwt.annotation
-   :no-inheritance-diagram:
-   :no-inherited-members:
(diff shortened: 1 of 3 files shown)


pkgw:
Ah, I was wondering about this. `annotation` with the lower-case "a" is the API file corresponding to the module `pywwt.annotation`. `Annotation` with the upper-case "A" is the API file corresponding to the Python class "Annotation" that is exported in the toplevel `pywwt` module. Even though it's defined in the `pywwt.annotation` module, the way that Sphinx's API autogeneration works, it wants to organize the class as belonging to the toplevel module.

I think the thing to do here is rename `pywwt.annotation.rst` since we have actual control over its filename, even though it is gross to have the filename not match the module name. The other file is auto-generated by the doc tools and it would be an API break to move or rename the class in order to get a different filename out. The module file could be renamed to something like `pywwt.annotation_.rst` with a corresponding update in the top-level `index.rst`. Not elegant, and it technically would break existing external links to our docs online, but I'd rather do that than break the Python API.

PR #331, opened 2022-08-08
Make it possible to customize app URL in Jupyter widget

pkgw:
### Overview

I think that this will help address WorldWideTelescope/wwt_kernel_data_relay#4.

### Checklist


pkgw:
Ah, looks like the switch to webkit means that we'll have to update all of the test images ...

PR #332, opened 2022-08-11
Allow time_decay attribute of layers to be serialized

Carifio24:
This PR allows the `time_decay` property of a table layer to be serialized - currently, using `_serialize_state` if the WWT widget is displaying a table layer will throw an error as it doesn't know how to serialize an astropy `Quantity`. The conversion is the same as is used in `_on_trait_change` [here](https://github.com/WorldWideTelescope/pywwt/blob/master/pywwt/layers.py#L1305).

The motivation is to allow `save_as_html_bundle` to work, with https://github.com/glue-viz/glue-wwt/pull/62 being a particular case where this functionality would be used. It would be nice to have the exported bundle use the research app, but I suspect that will take a nontrivial amount of effort.

### DIFF
--- a/pywwt/layers.py
+++ b/pywwt/layers.py
@@ -1319,6 +1319,8 @@ class TableLayer(HasTraits):
             wwt_name = trait.metadata.get("wwt")
             if wwt_name:
                 value = trait.get(self)
+                if wwt_name == "decay" and value is not None:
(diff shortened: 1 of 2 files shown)


pkgw:
There were a few CI errors here but they all looked spurious. I've relaunched them and hopefully this time they'll work.

PR #334, opened 2022-08-16
Allow toggling whether a table layer is selectable

Carifio24:
This PR adds the ability to select whether or not a table layer is selectable - basically, exposing the functionality from https://github.com/WorldWideTelescope/wwt-webgl-engine/pull/183. As in that PR, allowing client applications to make layers non-selectable should help performance when there are a lot of points, with `glue-wwt` being one example of an application that can benefit from having layers be non-selectable most of the time. To make this possible, I've added a boolean `selectable` traitlet to the table layer class. Since selectability is a property at the research app level, rather than of the engine layer, I've set `wwt=None` in the traitlet metadata.

### DIFF
--- a/pywwt/layers.py
+++ b/pywwt/layers.py
@@ -712,6 +712,12 @@ class TableLayer(HasTraits):
         "(:class:`~astropy.units.Quantity`)",
     ).tag(wwt="decay")
 
+    selectable = Bool(
+        True,
+        help="Whether sources in the layer "
+        "are selectable (`bool`)"
+    ).tag(wwt=None)
+
     # TODO: support:
     # xAxisColumn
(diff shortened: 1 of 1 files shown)


pkgw:
Looks like one of the macOS CI builds spuriously timed out, so I've restarted it. It would be nice to get test coverage of this, but seems like it might be hard ... any thoughts about whether/how that can be done?

Carifio24:
The only thing I can think of is using some of Qt's mouse functionality to move the pointer to a specific location where we know a source will be, toggling selectability, and seeing if the cursor shape changes. That feels like it'd be very brittle, though.

pkgw:
Yeah :-/

PR #335, opened 2022-08-19
Update for FITS TOAST and toasty 0.18

imbasimba:
- Replaced the parameters force_tan etc. with a TilingMethod enum
- Array/wcs input is now sent to TOASTY for tiling if the input is big enough
- Added option to set start level for TOAST creation

Requires https://github.com/WorldWideTelescope/toasty/pull/86

### DIFF
--- a/pywwt/layers.py
+++ b/pywwt/layers.py
@@ -278,8 +278,8 @@ class LayerManager(object):
             provided image.
         tiling_method : optional :class:`~toasty.TilingMethod`
             Can be used to force a specific tiling method, i.e. tiled
-            tangential projection, TOAST, HiPS, or even untiled. Defatults
-            to auto detection, which choses the most apropriate method.
+            tangential projection, TOAST, HiPS, or even untiled. Defaults
+            to auto-detection, which choses the most appropriate method.
         kwargs
             Additional keyword arguments can be used to set properties on the
             image layer or settings for the `toasty` tiling process. Common
--- a/pywwt/layers.py
+++ b/pywwt/layers.py
@@ -291,7 +291,8 @@ class LayerManager(object):
         kwargs
             Additional keyword arguments can be used to set properties on the
             image layer or settings for the `toasty` tiling process. Common
-            toasty settings include ``out_dir``, ``override``, and ``blankval``.
+            toasty settings include ``out_dir``, ``override``, ``blankval``,
+            and ``start``.
             See `toasty.tile_fits`.
 
         Returns
@@ -299,6 +300,10 @@ class LayerManager(object):
         layer : :class:`~pywwt.layers.ImageLayer` or a subclass thereof
         """
 
+        if isinstance(image, tuple):
+            image, wcs = image
+            from astropy.io.fits import PrimaryHDU
+            image = PrimaryHDU(image, wcs.to_header())
         if (
             isinstance(image, astropy.io.fits.ImageHDU)
             or isinstance(image, astropy.io.fits.PrimaryHDU)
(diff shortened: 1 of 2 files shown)


pywwt/layers.py
```diff
@@ -276,49 +276,48 @@ def add_image_layer(
...
             isinstance(image, astropy.io.fits.ImageHDU)
             or isinstance(image, astropy.io.fits.PrimaryHDU)
             or isinstance(image, astropy.io.fits.HDUList)
         ):
-            unused_file_name = self._get_unused_fits_name()
-            image.writeto(unused_file_name)
-            image = unused_file_name
+            filename = "fits_input_{}".format(
+                hash(str(image.data.tobytes()) + image.header.tostring())
+            )
```
pkgw, line 307: This bit of code reads in the *entire* image data array, which could be very large. For instance, in my DASCH test image, the `tobytes()` representation is about 3 gigabytes. We should probably limit this to the first 64k of data, or something like that.

Calling `str` on the `tobytes()` return value is also inefficient. In my sample, it returns a string that starts `b'\\x7f\\xf8\\x00\\x00...`. The string is 10 gigabytes!

Finally, I believe that the Python built-in `hash()` function now has some randomization built in so that hash values vary from one Python invocation to the next. So the filenames won't be stable across runs or Jupyter kernels, which seems to go against the intent of this code. This can be dealt with by using a stably-defined hash, such as:

```python
from hashlib import md5
d = md5()
d.update(image.data[:65536])
d.update(image.header.tostring().encode('utf-8'))
h = d.hexdigest()  # returns digest as a hex-encoded string
```

pywwt/layers.py
```diff
@@ -276,49 +276,48 @@ def add_image_layer(
...
             See `toasty.tile_fits`.
 
         Returns
         -------
         layer : :class:`~pywwt.layers.ImageLayer` or a subclass thereof
         """
 
+        if isinstance(image, tuple):
+            image, wcs = image
+            from astropy.io.fits import PrimaryHDU
```
pkgw, line 297: For consistency with the code below, this could use `astropy.io.fits.PrimaryHDU` instead of importing the name directly.

pywwt/layers.py
```diff
@@ -276,49 +276,48 @@ def add_image_layer(
...
-            image = unused_file_name
+            filename = "fits_input_{}".format(
+                hash(str(image.data.tobytes()) + image.header.tostring())
+            )
+            if not Path(filename).is_file():
+                image.writeto(filename)
+            image = filename
         if isinstance(image, str):
             image = [image]
         if isinstance(image, list):
```
pkgw, line 313: If I'm understanding the function logic correctly, at this point `image` *must* be a `list`. The `if` statement should be removed, and then the rest can be dedented.

pywwt/layers.py
```diff
@@ -713,9 +700,7 @@ class TableLayer(HasTraits):
     ).tag(wwt="decay")
 
     selectable = Bool(
-        True,
-        help="Whether sources in the layer "
-        "are selectable (`bool`)"
+        True, help="Whether sources in the layer " "are selectable (`bool`)"
```
pkgw, line 703: The autoformatter is a bit conservative here but the adjacent string literals can be combined into a single literal.

pkgw (changes requested):
Cool! There's one piece that I think needs some updating, but other than that this looks good, pending CI once a new Toasty is released.

imbasimba:
Agreed on all remarks!

pkgw (approved):
Looking good! Now we just need to update the minimum required version of Toasty, once a new release is made.

pkgw:
The new toasty, 0.18, is released. Various CI failures on the first build but they may all be spurious.

PR #339, opened 2022-10-19
Update interactive figure export to account for named colormaps

Carifio24:
This PR fixes a bug in the interactive figure export where named colormaps are not exported correctly. This is a pretty straightforward change, we just need to make sure to account for all of the relevant settings, which I did by mimicking the logic [here](https://github.com/WorldWideTelescope/pywwt/blob/master/pywwt/layers.py#L1126).

Also, I changed `"_colorMap"` to `"colorMap"` in the other two cases because using `"_colorMap"` doesn't call the setter correctly (since it's `set_colorMap` rather than `set__colorMap`).

### DIFF
--- a/pywwt/layers.py
+++ b/pywwt/layers.py
@@ -1371,11 +1371,20 @@ class TableLayer(HasTraits):
                 state["settings"][wwt_name] = value
 
         if self._uniform_color():
-            state["settings"]["_colorMap"] = 0
+            state["settings"]["colorMap"] = 0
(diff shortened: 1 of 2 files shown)


pkgw:
I think most of the test failures here are spurious, but flake8 has a complaint to address.

Carifio24:
I fixed the flake8 error, so the tests are all good now.

pkgw:
Great, thanks!

PR #342, opened 2022-12-01
Update the widget to research-app 0.12

pkgw:
### Overview

Here we'll try to update the widget to the latest release of the research app, 0.12, which contains the upgrade to Vue 3.

This may take some futzing because the new app contains the new engine which has some fixes to FITS rendering, so the test images may need updating.

### Checklist

- Changed code has some test coverage: N/A
- Changes in functionality documented: N/A


pkgw:
Hmm, discovered something interesting here. I'll copy/paste from my commit message:

> I recently updated the `builtin-image-sets.wtml` file to match the `imagesets6.xml` managed through the new-ish `wwt-core-catalogs` repository. I created `builtin-image-sets.wtml` back when we couldn't maintain the WWT servers in any reasonable way, so there was no mechanism to update the imageset list. Now we have much better systems for that and the `imagesets6` file was the one that was more recently maintained.
>
> Anyway, as part of that maintenance, the builtin SFD map switched from the virtualearth.net data source to the WWT tiles.aspx data source. And it turns out that they are slightly rotated relative to one another! Fortunately, the new version (tiles.aspx) is more correct. The imagery that we were using in our tests up until now used the older version, so we need to update the reference images here.
>
> Unrelated to this change, but it looks like we should check out the other datasets that are served off of virtualearth.net and see whether any of them also have astrometric issues.

PR #344, opened 2023-01-13
Fallback directory for image layers

Carifio24:
Currently, if the `image` argument of `add_image_layer`is something FITS-y, the image is written to disk to allow caching for Toasty. This causes a problem if the working directory is not writable. To work around that, this PR checks first if the working directory is writable. If not, the file is written to a temporary directory instead.

This temporary directory is stored as a member value on the layer manager. This way only one directory is created, and it can be cleaned up when the layer manager is destroyed.

### DIFF
--- a/pywwt/layers.py
+++ b/pywwt/layers.py
@@ -240,18 +240,23 @@ class LayerManager(object):
         self._parent = parent
         self._tmpdir = None
 
-    def __del__(self):
-        if self._tmpdir is not None:
-            shutil.rmtree(self._tmpdir)
-
-    def _toasty_input_filepath(self, hex_string):
+    def _write_image_for_toasty(self, image, m):
+        hex_string = m.hexdigest()
         filename = "toasty_input_{}.fits".format(hex_string)
-        if access(getcwd(), W_OK):
+        if Path(filename).is_file():
             return filename
 
-        if self._tmpdir is None:
-            self._tmpdir = tempfile.mkdtemp()
-        return path.join(self._tmpdir, filename)
+        try:
+            image.writeto(filename)
+            return filename
+        except OSError:
+            if self._tmpdir is None:
+                self._tmpdir = tempfile.TemporaryDirectory()
+            filepath = path.join(self._tmpdir.name, filename)
+
+            if not Path(filepath).is_file():
+                image.writeto(filepath)
+            return filepath
 
     def add_image_layer(
         self,
@@ -343,13 +348,7 @@ class LayerManager(object):
                         m.update(hdu.data[:65536])
                         m.update(hdu.header.tostring().encode("utf-8"))
 
-                hex_string = m.hexdigest()
-                filepath = self._toasty_input_filepath(hex_string)
-
-                if not Path(filepath).is_file():
-                    image.writeto(filepath)
-
-                image = filepath
+                image = self._write_image_for_toasty(image, m)
 
         if isinstance(image, astropy.io.fits.ImageHDU) or isinstance(
             image, astropy.io.fits.PrimaryHDU
@@ -359,13 +358,7 @@ class LayerManager(object):
             m = md5()
             m.update(image.data[:65536])
             m.update(image.header.tostring().encode("utf-8"))
-            hex_string = m.hexdigest()
-            filepath = self._toasty_input_filepath(hex_string)
-
-            if not Path(filepath).is_file():
-                image.writeto(filepath)
-
-            image = filepath
+            image = self._write_image_for_toasty(image, m)
 
         if isinstance(image, str):
             image = [image]
--- a/pywwt/layers.py
+++ b/pywwt/layers.py
@@ -240,14 +240,16 @@ class LayerManager(object):
         self._parent = parent
         self._tmpdir = None
 
-    def _write_image_for_toasty(self, image, m):
+    def _toasty_filename(self, image, hdu_index=None):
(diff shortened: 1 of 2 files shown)


pkgw (changes requested):
I'll admit that this is the kind of thing that is barely ever going to matter in daily use, but the current approach in this PR has a flaw. "Testing whether you can do something, then attempting to do the thing" introduces a classic race condition into the filesystem interactions — the filesystem can change between the test and the operation. I don't see how it could be relevant here, but this is an easy way to introduce security vulnerabilities into operations that interact with the filesystem.

The only Correct way to do this kind of stuff is to just attempt the operation, and handle the error if/when it occurs. In this case it would require some code rearrangement but should basically come down to a try/except block. I think it will be worth adding a test that triggers the permission-denied behavior since it's the kind of thing that won't come up in many applications but will be relevant to some of them (e.g. the Glue DMG stuff).

Also, the `tempfile.TemporaryDirectory` class has some support for automatic cleanup of the directory, and it's probably better to tie into its implementation rather than rolling our own, although I don't know if it does anything cleverer than what we have here.

Carifio24:
@pkgw Good point about the race condition. I've changed the implementation to attempt the initial write before falling back to the temporary directory. To do this, I've replaced `_toasty_input_filepath` with `_write_image_for_toasty`. This method returns the filepath that was ultimately used, since that's used later in `add_image_layer`. This method also follows the existing `add_image_layer` logic of not writing if the filepath also exists.

I've changed to using the `TemporaryDirectory` class as you've suggested. Per the `tempfile` [docs](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory), this directory is cleaned up upon destruction of the directory object (in our case, when the layer manager is destroyed), which removes the need to handle this in `__del__`. (Some local testing agreed with this.)

I'm not sure yet how to specifically trigger this in a test - will keep thinking about it. I also was unsure whether we want to catch any possible errors in the temporary directory write - I don't know that there's more we can do at that point, so I've left it unhandled.

Carifio24:
I've done a bit more refactoring and added a test. The refactoring was partially to help with testability. I also put the `Path(filename).is_file()` check into the try block, since that can throw a `PermissionError` (a subclass of `OSError`).

The test uses an `HDUList` as input, which is the type of case that was causing the issue in glue. The test (temporarily) moves the current working directory to a temporary directory and first checks the default case. It then makes that directory non-writable to check the fallback case, and finally restores the directory's previous permissions.

pkgw (approved):
Looks good now! I pushed some small polish changes but I think this is all set now.

pkgw:
I need to close and reopen this to work through a transient Azure Pipelines error.

pkgw:
Ah, it seems that this is not some transient issue after all. It looks like the new Azure Pipelines builder image has some issue with its built-in Conda setup that's not allowing basic installs to work. In my most recent debugging commit a very basic `conda install -y mamba` is failing due to conflicts.

The most recent successful build is image version `20221125.1`, while we're now on `20230118.1`.

I'll try pushing a few other commits to dig around ... this may be related to our use of conda-forge.

pkgw:
OK, it looks like it helps the CI if we just create a new environment with conda-forge rather than modifying the base one.

But, the tests are showing a bug if the argument to `add_image_layer()` is a string. It looks like the processing chain needs to get some logic to only do the MD5 stuff if the input is an HDU or HDUList. It might also help clarify things to reorder the code since there's a pretty straightforward sequence of operations: `add_image_layer` → `_write_image_for_toasty` → `_toasty_filename` → `_image_md5`.

pkgw:
Some of these test failures look like intermittent timeouts, but it's something different on Windows. My guess is that we're not actually making the test directory read-only on Windows, or at least what we're doing doesn't make it an error to write the new file, so the tempdir doesn't actually get used?

Carifio24:
I tried this out on my Windows partition, and that is indeed the issue. It seems that `chmod` works on files, but not directories, on Windows. Since we assume that nothing needs to be done if the filepath already exists (to save the unnecessary write), we can't just make a file with the appropriate name and make it read-only.

I found [this SO thread](https://stackoverflow.com/questions/26465546/how-to-authorize-deny-write-access-to-a-directory-on-windows-using-python), but that seems like too much just for this test case on one platform. I'll keep thinking about a workaround.

Carifio24:
So I came up with a somewhat kludgy workaround. What I have the test do on Windows is create a directory with the exact name as what our first write attempt would be. This will trigger the write attempt (since there's no file with the desired name), but will give a `PermissionError` since the write will fail (and thus lead to the fallback behavior).

pkgw:
Nice solution! And I think it's good to get that extra bit of cross-platform coverage in the tests.

pywwt/layers.py
```diff
@@ -302,7 +303,71 @@ def add_image_layer(
         # has previously been processed. As usual, the logic is a bit
         # complicated since we aim to allow the user to give us just about
         # anything as an input.
+        if any(isinstance(image, fitsy_type) for fitsy_type in [
+            astropy.io.fits.HDUList,
+            astropy.io.fits.ImageHDU,
+            astropy.io.fits.PrimaryHDU
+        ]):
```
pkgw, line 310: Oh, not worth editing things, but IIRC `isinstance(image, (HDUList, ImageHDU, PrimaryHDU))` actually has these exact semantics.

Carifio24: Ah, yep, I forgot that `isinstance` can take a tuple

PR #345, opened 2023-01-24
Serialize settings for size attributes

Carifio24:
Serialization of size attributes currently doesn't work. We have a special column name (`SIZE_COLUMN_NAME`) designated for storing sizes in the serialization, but that column name is never actually used in the layer's data table. This PR updates the serialized state to match the settings messages that are dispatched [here](https://github.com/WorldWideTelescope/pywwt/blob/master/pywwt/layers.py#L1061).

I'm not sure how often serialization is used, but one intended use case would be https://github.com/glue-viz/glue-wwt/pull/62, as using size attributes is a common operation in glue.


pkgw:
Hmm. It's been a little while since I've thought about this aspect of things, so I want to be careful, but I think that this is correct. The key is that pywwt used to compute the sizes *for* WWT, and used the magic column name for those computed sizes. But in 9c358cf64a901c3ee2a878e55e58ac544e366bfa we started relying on built-in support for the engine and I think that we stopped needing to use the special column. Does that jibe with your understanding?

Carifio24:
Yep! That's how I had understood things as well.

Carifio24:
The CI failure looks like a transient timeout thing? But I suspect it would've needed another run anyways to pick up the new CI updates

pkgw:
I think you'll need to rebase this onto the latest `master` (or merge it in to your branch) for it to pick up the CI fixes.

Carifio24:
Just rebased

PR #349, opened 2023-03-01
Expose roll angle

Carifio24:
This PR follows up on https://github.com/WorldWideTelescope/wwt-webgl-engine/pull/231 to expose the roll angle Python-side. This includes attaching the roll angle to the pywwt widget, allowing users to ask for it (`get_roll`), and allowing roll angles to be used in `center_on_coordinates`. There I opted to make the default value `roll=None` (and only include it in the message if it isn't) to match the research-app-messages API, where the parameter is optional. Note that this involves bumping the research-app version to account for the `ViewStateMessage` change.

My only two questions are:
* The `SkyCoord` returned by `get_center` has a `roll` member value. Should the roll angle be exposed there as well/instead?
* Is it worth adding a test for this? The obvious thing would be to add a stage in the Qt widget's `test_full` that has a nontrivial roll angle.


Carifio24:
I agree that an image test would be nice (maybe with a value-roundtripping test as well). Have we typically used any set methodology for generating the reference images? (e.g. from some particular matrix of OSes and Qt versions)

pkgw:
@Carifio24 It's pretty tedious, unfortunately. You can copy an existing image over as a placeholder, then submit the new testcase to CI, then download the artifacts resulting from all the failed tests and use some of the support tooling (see the README) to add any other variations that will help things pass. The images generated on my Linux laptop are different than the ones that get generated during CI so I add further images so that I can test locally without getting errors, even though they're not necessary for CI to pass.

Carifio24:
@pkgw I've added a non-image test, but I'm planning on adding an image one as well.

pkgw:
Um, I'm not sure why the robot suddenly got involved, but you can go ahead and agree with it with a company of "Harvard University", I guess.

Carifio24:
@dotnet-policy-service agree company="Harvard University"

pkgw:
Awesome, thanks!

Carifio24:
In case it's ever useful information: this PR added four reference images for the new step in `test_full`: `a` is my local (Ubuntu 22.04) output, and `b`, `c`, and `d` are the Linux, Mac, and Windows CI outputs (with each OS, the images from different Python versions were identical, per `filecmp`).

PR #354, opened 2023-06-28
Expose grid text and additional constellation settings

Carifio24:
This PR exposes the ability to toggle the grid text for the four grids that are exposed (equatorial, ecliptic, alt/az, and galactic), as well as the ability to toggle constellation labels and pictures. One intended use case is `glue-wwt` (i.e. as an extension to what's currently in https://github.com/glue-viz/glue-wwt/pull/97).

The implementation for some of these properties already existed but were commented out. There are some notes about the constellation properties in the commit where they were commented out [here](https://github.com/WorldWideTelescope/pywwt/commit/761dc72f96a7de4976431401e0db56f411b1096a). In my testing both of these properties work fine (and I don't see any errors coming up while scrolling with the constellation pictures enabled), so my suspicion is that these comments relate to an older (pre-research app) pywwt implementation. Regardless, all of these properties work fine in my local testing.

I decided to name the member value corresponding to the equatorial grid text as just `grid_text` for consistency with the `grid` member value. It's worth noting that the text for a grid won't draw unless the grid is displayed - i.e. the value of `alt_az_text` has no effect unless `alt_az_grid` is `True`. This is consistent with the internal engine settings.

If we're okay with the implementation here, I'll update the image testing to take these settings into account.


Carifio24:
Okay, I think this should be all set now. The only thing to note is that I increased the wait time for the new test step that checks some of the constellation settings - loading the constellation pictures needed some extra time, at least when testing.

pkgw:
Great, thanks!

PR #355, opened 2023-07-03
Expose various color settings and precession chart

Carifio24:
This PR exposes the color settings added in https://github.com/WorldWideTelescope/wwt-webgl-engine/pull/258 to pywwt, as well as exposing the ability to toggle the precession chart. This required bumping the research app to the newest version. I've added two new steps to the full widget test to check that these settings are working correctly.

As for the naming, I decided to go with `<gridtype>_grid_color` (e.g. `alt_az_grid_color`) rather than `<gridtype>_color` because this would be ambiguous in the case of the ecliptic vs. the ecliptic grid.


pywwt/core.py
```diff
@@ -693,6 +706,10 @@ def instruments(self):
         False, help="Whether to show a grid relative to the " "galactic plane (`bool`)"
     ).tag(wwt="showGalacticGrid", wwt_reset=True)
 
+    galactic_grid_color = Color(
+        "cyan", help="The color of the equatorial grid " "(`str` or `tuple`)"
```
pkgw, line 710: Copy/paste-o in the help string here.

pkgw (commented):
Cool! Besides the test errors and the comment below, this looks good to me.

Carifio24:
I've fixed the copy/paste-o and some small codestyle issues. Not sure what happened with the image test - I must've missed committing one of the CI image files before, since it's fine after adding the last Windows output image.

PR #356, opened 2023-07-20
Update ecliptic grid color default

Carifio24:
With all the colors getting exposed in #355, I accidentally gave the ecliptic grid the wrong starting color (maybe got it mixed up with the ecliptic color?) - it should be [green](https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/master/engine/wwtlib/Settings.cs#L358), rather than blue, to match the engine. This PR fixes that.

### DIFF
--- a/.check_enabled.py
+++ b/.check_enabled.py
@@ -2,22 +2,32 @@ import logging
 import sys
 
 from jupyterlab.commands import get_app_info
-from notebook.nbextensions import validate_nbextension
-from notebook.serverextensions import validate_serverextension
+
+try:
+    from nbclassic.nbextensions import validate_nbextension
+except ImportError:
+    # `notebook` <= 6
+    from notebook.nbextensions import validate_nbextension
+
+try:
+    from nbclassic.serverextensions import validate_serverextension
+except ImportError:
+    # `notebook` <= 6
+    from notebook.serverextensions import validate_serverextension
 
 # If there's a problem and we don't provide this, the validate function crashes :-(
-logger = logging.getLogger('')
+logger = logging.getLogger("")
 
-if validate_nbextension('pywwt/extension', logger=logger) != []:
+if validate_nbextension("pywwt/extension", logger=logger) != []:
     print("Issue detected with nbextension")
     sys.exit(1)
 
 info = get_app_info()
 
-if 'pywwt' not in info['extensions'] or 'pywwt' in info['disabled']:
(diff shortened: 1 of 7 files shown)


Carifio24:
The CI failures are unrelated, and seem to be an issue with the Jupyter `notebook` package - version 7.0.0 of that package released yesterday.

There's a [migration guide](https://jupyter-notebook.readthedocs.io/en/stable/migrating/frontend-extensions.html) for Notebook 7, but updating sounds like it could be a real task - maybe we should pin for now?

pkgw:
Well, Notebook 7 is a pretty major update, but we don't actually use the `notebook` module much — just some infrastructure references in `jupyter_server` and `jupyter_relay`, for the main package. It looks like the build errors are mostly the `.check_enabled.py` helper script, which hopefully should be easy to update to be compatible with both major versions.

Carifio24:
So from what I can see, we're still getting errors because `jupyter_contrib_core` (pulled in through `jupyter_contrib_extensions` [here](https://github.com/WorldWideTelescope/pywwt/blob/master/ci/azure-build-and-test.yml#L147))  hasn't been updated for Notebook 7 (i.e. they need to do what you just added in the last commit). See e.g. [here](https://github.com/Jupyter-contrib/jupyter_contrib_core/blob/master/src/jupyter_contrib_core/notebook_compat/nbextensions.py) (their `_compat.nbextensions` also imports `notebook` directly).

pkgw:
Things seemed to be working for me with `notebook >= 7` and without that package installed, so I think the question is, can we perhaps now avoid installing it altogether? Maybe we only need one of its dependencies?

Carifio24:
That explicit installation inside the CI workflow is the only reference I can find to `jupyter_contrib_nbextensions`. It was added in [this commit](https://github.com/WorldWideTelescope/pywwt/commit/46cc64538c19de7536812b4c67b8cd5bf463011a), so I'm assuming the `jupyter nbextension list` command was failing without it? (the logs are no longer around). From looking at their list of included extensions, it's not immediately clear to me why that would be, but yeah, maybe we could find the relevant extension and only install that.

pkgw:
Current build failure is a conda problem unrelated to this PR; I think it will be resolved over the next day or so. (I think the current release of conda needs to be yanked and replaced.)

pkgw:
Cool, yes, looks like this problem is dealt with now. The Linux py310 build failed with a segfault while running the test suite (!) but hopefully that is a once-in-a-blue-moon problem ...

PR #368, opened 2023-11-12
Compatibility with ipywidgets 8.x

Carifio24:
This PR aims to make the pywwt widget compatible with ipywidgets 8.x, which currently seems to not be the case. I was trying to use the Jupyter widget in a glue-wwt context together with some of the new widgets from ipywidgets 8 but kept getting the following error:

```
Model class 'WWTModel' from module 'pywwt' is loaded but can not be instantiated
TypeError: r._deserialize_state is not a function
    at f._make_model (http://localhost:8889/lab/extensions/@jupyter-widgets/jupyterlab-manager/static/150.b0e841b75317744a7595.js?v=b0e841b75317744a7595:1:7933)
```

After some investigation, it seems that we need to update our widget JS to use ES6 class notation (see [here](https://ipywidgets.readthedocs.io/en/latest/migration_guides.html#backbone-extend)), which is the content of this PR. With these changes, the widget works for me locally with both ipywidgets 7.x and 8.x.

### DIFF
--- a/frontend/lib/widget.js
+++ b/frontend/lib/widget.js
@@ -67,21 +67,24 @@ var version = require('./index').version;
 // Which is a hairy busines, because not only does our one widget potentially
 // have multiple views, but there are also potentially multiple active widgets,
 // and we'll see all of their messages.
-var WWTModel = widgets.DOMWidgetModel.extend({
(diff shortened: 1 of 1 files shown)


pkgw:
Thanks for tracking this down!

PR #373, opened 2024-04-11
Add DECaPS2 to default pywwt surveys

Carifio24:
I got a request to expose the DECaPS2 RGB survey to `glue-wwt`. The most straightforward way to do that is to add it to the default `pywwt` surveys, and then `pywwt` users can access it as well. The WTML tag here is taken verbatim from the default webclient surveys.

PR #374, opened 2024-04-12
Add DECaPS2 to default pywwt surveys

Carifio24:
Re-submission of #373.


pkgw:
As worked out in the previous PR: note that this PR is correctly submitted to the `gh-pages` branch, which is why the usual Azure Pipelines CI isn't running — no pipelines YAML files here. The ReadTheDocs CI should not be running either, but their system doesn't appear clever enough to recognize that. That CI "failure" should be ignored here.

PR #375, opened 2024-04-18
Allow refreshing tile cache

Carifio24:
This PR exposes the cache-refreshing functionality of the research app to pywwt. We have a use case for this in CosmicDS (not urgent, but it was easy so I figured I'd get the ball rolling), but I think having the option available is generally useful. I did add a cautionary note in my docstring that this should really only be used when it's needed, though.

If we want a test, I'm not really sure what a good one would be - I don't see a straightforward way to use the image comparison method here.


Carifio24:
It looks like there's a pin in the CI (`pin-1`) that's requiring Python 3.12(?). I don't understand why conda would use Python 3.12 when we set the Python version [here](https://github.com/WorldWideTelescope/pywwt/blob/master/ci/azure-build-and-test.yml#L50), though.

Carifio24:
@pkgw Ok, I think I have a beat on this now. So while we specify our Python version where I had linked above, the offending step is actually before that, in the setup. There we're installing `brotlipy` using the base environment (the comment says it's for a Windows workaround), which is now using Python 3.12 by default.

According to the [runner info](https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20240422.1), it's using Miniconda 24.3.0. Miniconda added support for Python 3.12 in 24.1.2. The last CI prior to that was [using Miniconda 23.10.0](https://github.com/actions/runner-images/blob/ubuntu22/20231211.1/images/ubuntu/Ubuntu2204-Readme.md), so no Python 3.12, and thus not surprising that we didn't see this issue.

I'll admit I don't know exactly what we use `brotlipy` for, as it only seems to be referenced in that part of the CI. The Brotli project itself offers [a `brotli` Python package](https://github.com/google/brotli?tab=readme-ov-file#python), whereas `brotlipy` seems to have been become/been replaced by [`brotlicffi`](https://github.com/python-hyper/brotlicffi). Maybe it's time to upgrade?

pkgw:
We wouldn't use Brotli(py) directly. In my own system, I see that urllib3 depends on it, though, so it probably comes in as a transitive dependency somewhere in the web stack (which makes sense, since Brotli is used as a compression format by modern web servers).

I'm afraid that I don't remember exactly what was going on that inspired me to add in the install step — probably some kind of missing library or something that prevented Conda from running. I've just pushed a commit to just remove the install since it was just a workaround to begin with.

pkgw:
The test that involves the `sky_layers` image starts by setting the foreground and background to the "Black Sky Background" imagery ([here][1]), so it's incorrect for stars to be showing up in the test image. It's not obvious to me why the behavior would be changing here. Clearly this PR shouldn't affect that, and I can't think of other recent changes that would either.

[1]: https://github.com/WorldWideTelescope/pywwt/blob/master/pywwt/tests/test_layers.py#L354-L355

To my eye, the other image changes seem to be stemming from the background map not having tiled in to its full resolution just yet. Fetching those tiles is something that should be pretty reliable. It's possible that this could be worked around by increasing the delays in the tests, to give the engine more time to load images, but it might be good to double-check that this isn't indicating that our webservices are having some kind of issue.

PR #376, opened 2024-05-14
Make qtpy dependency optional

Carifio24:
Downstream in `glue-wwt` we're trying to make `qtpy` be an optional dependency (https://github.com/glue-viz/glue-wwt/issues/111) so that it doesn't need to be installed to use the WWT Jupyter viewer. To help make that possible, this PR makes `qtpy` an optional dependency of `pywwt`, moving it to the `qt` and `qt6` install options.

The [documentation](https://github.com/WorldWideTelescope/pywwt/blob/master/docs/installation.rst?plain=1#L221) already makes it sound like `qtpy` is an optional dependency, so I don't think it needs to be changed at all?

### DIFF
--- a/ci/azure-build-and-test.yml
+++ b/ci/azure-build-and-test.yml
@@ -199,7 +199,7 @@ jobs:
       set -x
       \conda install -y pytest-cov pytest-remotedata pytest-timeout
       python setup.py build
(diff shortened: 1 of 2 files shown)


Carifio24:
CI failures are the same as in #375; I can rebase once we finish sorting that out.

pkgw:
Looks like you'll need to rebase to make ReadTheDocs happy.

PR #378, opened 2024-12-29
Deal with integer FITS files with BLANK keywords

pkgw:
### Overview

It turns out that `reproject` has a bug in dealing with integer FITS images that have BLANK keywords. This affects the cutouts images processed by daschlab. This should be fixed upstream, but that might take a while, so in the meantime let's work around it.

We also refresh the API docs and add a script to help make sure they get updated regularly.

### Checklist

No test coverage, sorry :-(


pkgw:
Upstream bug report is https://github.com/astropy/reproject/issues/475 .

pkgw:
Also adding multi-wcs support here since it turns out to be a small change.

PR #380, opened 2025-02-27
Replace pipes with shlex in setupbase

Carifio24:
We currently have a reference to the `pipes` package in our `setupbase` utilities, has been deprecated since Python 3.11 and has now been removed in 3.13. Our only need for `pipes` is a single use of `pipes.quote`, which was [already being imported from `shlex`](https://github.com/python/cpython/blob/3.12/Lib/pipes.py#L66), so this PR replaces it with `shlex.quote`.

FWIW, I noticed this issue in a downstream `glue-wwt` [CI job](https://github.com/glue-viz/glue-wwt/actions/runs/13123865470/job/37950729027).

### DIFF
--- a/setupbase.py
+++ b/setupbase.py
@@ -14,7 +14,6 @@ from os.path import join as pjoin
 import io
 import os
 import functools
-import pipes
 import re
 import shlex
 import subprocess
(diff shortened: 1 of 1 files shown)


Carifio24:
Hmm, not immediately sure what the segfault in the CI builds are from - this really shouldn't have changed anything...

PR #381, opened 2025-03-12
Fix OpenGL issues in Linux CI

Carifio24:
Currently the Python 3.9 and 3.10 jobs in the Linux CI are failing with segmentation faults, which I think is related to their not being able to recognize OpenGL. This PR is an attempt to think that. I suspect that this will require a good bit of iteration.


pkgw:
Looks good ...

PR #382, opened 2025-03-26
Allow specifying URL of surveys to initially load in

Carifio24:
### Overview

This PR makes it possible to pass in a URL to the widget specifying the location of a WTML with initial surveys to load. If no URL is given (or the given URL is an empty string), the fallback behavior is to use the default pywwt survey WTML. I imagine most users won't have any need for this, but I have a downstream application in mind, and it might be nice for users who know what they're doing.

### Checklist

- [x] Changed code has some test coverage (or justify why not)
- [x] Changes in functionality documented (or justify why not)

I think a good test would be to try loading in some other WTML collection and verifying that properties like `_available_layers` match the expected content afterwards. @pkgw is there another WTML on the WWT servers (or somewhere equally reliable) that would be appropriate for this?


Carifio24:
> The current version is attached to the pywwt GitHub Pages content (gh-pages branch of this repo) so you could add something there.

I've added a small `test_surveys.xml` file into the `gh-pages` branch, as well as a simple test that makes sure that basic imagery properties are as expected when creating the widget using that XML for the surveys.

> In case we've never discussed this ... you might be wondering why pywwt has its own special list of surveys attached to its GitHub Pages site, anyway.

I actually don't think we've ever discussed this - or if we have, I'm just not remembering it. Good to know though! Seems like that was the most reasonable approach at the time.

PR #385, opened 2025-05-22
Try and fix ctypes DLL issue in Windows CI

Carifio24:
The Windows CI is failing due to some issue with importing `_ctypes`  when reactivating the conda environment. This PR is an attempt to fix that.

From what I've gathered from reading online, it sounds like this might be an issue with the `libffi` DLL being missing. Let's see if that's the case.


Carifio24:
Closing this as the CI issues were fixed in #391.

PR #388, opened 2025-07-22
Accept Anaconda ToS for necessary channels

Carifio24:
We're getting an error in the CI due to needing to accept some Anaconda Terms of Service for a couple of channels. Reading over the Anaconda documentation, I think we quality for free for any reasonable definition of "we" (thinking in particular of NumFOCUS, see [here](https://www.anaconda.com/legal/terms/non-profit-research#elementor-toc__heading-anchor-1)). So this PR updates our CI script to accept the necessary ToS.


Carifio24:
The failure here is what #385 is trying to solve, so I think this is good now.

